Issue #12, opened 2025-12-17
Publish a minimal documentation website


niksirbi:
Minima requirements:
- [ ] installation instructions (for users and dev), probably with `uv`
- [ ] API index
- [ ] Sphinx gallery

PR #19, opened 2026-01-23
Add function to update image IDs in COCO JSON and add DLC test fixtures

sfmig:
## Description

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**
We agreed the following re the IDs in the frame labels COCO files:
- the image ID refers to the frame (0-indexed) in the video the file was extracted from
- the annotation and category IDs are 1-indexed (updated)

The annotation and category ID are 1-indexed keeping in line with sleap-io output and because `category_id`=0 is often reserved for the background class. Note that in principle, IDs in a COCO only need to be consistent within the file, but it may be that other frameworks make assumptions about them:
- For example, faster-rcnn in pytorch assumes this (see [here](https://github.com/pytorch/vision/blob/ef2920cc80bac61282b3b19a775b3c33de4e7551/torchvision/models/detection/roi_heads.py#L36), and the original comment about it [here](https://www.reddit.com/r/deeplearning/comments/r0f9rc/quick_question_can_there_be_no_0_label_in_the/)), 
- It seems [Google's mediapipe API too](https://discuss.roboflow.com/t/managing-category-id-0-in-coco-json-export/6994) 
- I haven't found this explicitly said in the original [COCO format description](https://cocodataset.org/#format-data) tho.

**What does this PR do?**
- It updates **only the image IDs** in the loaded COCO dict following the convention above before exporting the file. 
- It adds a check to verify the input annotations contain a single video only
- It adds tests

It also adds tests for `annotations_to_coco` and the fixtures used in PR #20 

**Questions for the reviewer(s)**
- If we decide that for consistency we do want category and annotation IDs to be 0-indexed, we can recover the relevant commits from this PR (including tests). 
	- In that case, I assigned the new category and annotation IDs in the order they appear in the file (i.e. with `enumerate`). Maybe it make more sense to assign the new IDs based on the sorted list of old IDs.
	
- I made each update function a "pure function" (I deep copy the input dictionary and return the modified one). Initially I was doing it in place but I think that could lead to tricky bugs... Let me know thoughts.

## References
PR #24 

Tests pass locally and in CI

## Is this a breaking change?

No.

Docstrings have been updated. We should probably include definitions for the keys in the COCO file as part of #12 

## Checklist:

### DIFF
--- a/poseinterface/io.py
+++ b/poseinterface/io.py
@@ -121,10 +121,7 @@ def _update_image_ids(input_data: dict) -> dict:
     for img in data["images"]:
         # map old image_id to new image_id
         old_img_id = img["id"]
-        new_img_id = _extract_frame_number(
-            img["file_name"],
-            POSEINTERFACE_FRAME_REGEXP,
-        )
+        new_img_id = _extract_frame_number(img["file_name"])
         old_to_new_id[old_img_id] = new_img_id
 
         # update image_id in images list
@@ -144,7 +141,9 @@ def _update_image_ids(input_data: dict) -> dict:
     return data
 
 
-def _extract_frame_number(filename: str, frame_regexp: str) -> int | None:
+def _extract_frame_number(
+    filename: str, frame_regexp: str = POSEINTERFACE_FRAME_REGEXP
+) -> int | None:
     """Extract the frame number in the input filename.
 
     If no frame number is found, returns None.
--- a/tests/test_unit/test_io.py
+++ b/tests/test_unit/test_io.py
@@ -162,17 +162,16 @@ def test_update_image_ids_duplicate_ids():
 @pytest.mark.parametrize(
     "filename, frame_regexp, expected_image_id",
     [
-        # default regexp
-        ("img0000.png", r"img(0\d*)", 0),
+        ("img0000.png", r"img(\d*)", 0),
         ("img0234.png", r"img(0\d*)", 234),
-        # custom regexp matching "frame-0NNN"
         (
             "sub-M708149_ses-20200317_view-topdown_frame-00000.png",
-            r"frame-(0\d*)",
+            POSEINTERFACE_FRAME_REGEXP,
             0,
         ),
-        ("frame-0234", r"frame-(0\d*)", 234),
-        ("frame-0234abcd", r"frame-(0\d*)", 234),
+        ("frame-234", POSEINTERFACE_FRAME_REGEXP, 234),
+        ("frame-0234", POSEINTERFACE_FRAME_REGEXP, 234),
+        ("frame-0234abcd", POSEINTERFACE_FRAME_REGEXP, 234),
     ],
 )
 def test_extract_frame_number(filename, frame_regexp, expected_image_id):


poseinterface/io.py
```diff
@@ -1,7 +1,11 @@
+import copy
+import json
+import re
 from pathlib import Path
 
 import sleap_io as sio
-from sleap_io.io import dlc
+from sleap_io.io.coco import convert_labels
```
lochhh: coco.convert_labels may be clearer as each format has its own convert_labels

poseinterface/io.py
```diff
@@ -17,22 +21,30 @@
...
-    """Export annotations file to COCO format.
+    """Export annotations file to COCO format for poseinterface.
 
     Parameters
     ----------
     input_path : pathlib.Path
         Path to the input annotations file.
     output_json_path : pathlib.Path
         Path to save the output COCO JSON file.
+    frame_regexp: str, optional
```
lochhh: how about dropping this and just extract the last group of digits matching `r"(0\d*)"`

tests/test_unit/test_io.py
```diff
@@ -0,0 +1,191 @@
...
+        (lf("dlc_multi_index_in_project_root"), "dlc"),
+    ],
+)
+def test_annotations_to_coco_invalid(
+    mock_load_file,
+    mock_is_dlc_file,
+    input_file,
+    error_message,
+    tmp_path,
+    request,
```
lochhh: unused?
```suggestion
```

tests/conftest.py
```diff
@@ -0,0 +1,133 @@
...
+
+    return csv_path_in_project
+
+
+@pytest.fixture
+def dlc_single_index_in_video_folder(tmp_path: Path) -> Path:
+    """Mock DLC project: single-index CSV in video folder (same as frames)."""
+    return create_dlc_project(
+        tmp_path, "single-index", csv_location="video_folder"
+    )
```
lochhh: These fixtures should be created once at the start of the entire test run and reused across all tests:
```suggestion
@pytest.fixture(scope="session")
def dlc_single_index_in_video_folder(
    tmp_path_factory: pytest.TempPathFactory,
) -> Path:
    """Mock DLC project: single-index CSV in video folder (same as frames)."""
    return create_dlc_project(
        tmp_path_factory.mktemp("dlc_single_video"),
        "single-index",
        csv_location="video_folder",
    )
```

poseinterface/io.py
sfmig: I am not sure if they are always in this format or preceded by at least one zero in all pose estimation frameworks, so was trying to be flexible

lochhh: `re.findall(r"\d+", "sub38_frame0001.png")[-1]` would work

tests/conftest.py
sfmig: why not on demand? 🤔 

Right now they are only used in the integration test. Maybe if/when they are needed further we can expand it?

lochhh: Not sure i understand... the proposed change here is to create these only once, rather than calling create_dlc_project every single time the fixture is called

lochhh: Ah yeah you're right. My PR #26 is waiting on these fixtures 😆

PR #26, opened 2026-02-06
Refactor `annotations_to_coco`

### LINKED ISSUE #23: Helper function to generate `coco_image_filenames` for `annotations_to_coco`
**Is your feature request related to a problem? Please describe.**
Currently we need to manually supply `coco_image_filenames` to `annotations_to_coco` to set the `file_name` value in the COCO json output. 

**Describe the solution you'd like**
For framelabels.json the file_name should be:
```
f"sub-{subject_id}_ses-{session_id}_cam-{cam_id}_frame-{frame_number}.{frame_file_ext}"
```
For cliplabels.json:
```
f"sub-{subject_id}_ses-{session_id}_cam-{cam_id}_frame-{frame_number}"
```
We could have a helper function that takes `input_path`, `subject_id`, `session_id`, `cam_id` to generate this list of filenames to pass to `annotations_to_coco`.

lochhh:
## Description

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**
Closes #23 

**What does this PR do?**
- Adds a helper function to auto-generate the list of image file names to be passed to coco.convert_labels
  - the file names are 0-padded to match the width of the largest frame_idx in the dataset 
- Refactors `annotations_to_coco` to generate cliplabels or framelabels.json with the correct file names and image_ids that conform to poseinterface specs.
  ```python
    coco_json = annotations_to_coco(
        input_csv,
        output_dir,
        sub_id="sub123",
        ses_id="ses123",
        cam_id="cam123",
        mode="clip", # or "frame"
    )
  ``` 
- Adds a minimal .slp file with 5 labelled frames for testing
- Reorganises test data into software-specific dirs
- Adds cliplabels.json for the sleap test data and framelabels.json for multi-index DLC test data 
- Adds relevant tests
- Updates the example accordingly

## References
Incorporates the update id function introduced in #19 into annotations_to_coco 

The cliplabels.json and framelabels.json are generated using annotations_to_coco, useful to manually verify that the function works as expected. These are also then used in tests.

## Checklist:


poseinterface/io.py
```diff
@@ -28,33 +29,43 @@ def annotations_to_coco(
     input_path: Path,
     output_json_path: Path,
     *,
-    coco_image_filenames: str | list[str] | None = None,
-    coco_visibility_encoding: str = "ternary",
+    sub_id: str,
+    ses_id: str,
+    cam_id: str,
+    mode: Literal["clip", "frame"] = "frame",
 ) -> Path:
```
sfmig, line 38: Should we rename to `annotations_to_poseinterface` ?

To make it clear that COCO format and poseinterface is not the same (poseinterface is a "type" of COCO format)

poseinterface/io.py
```diff
@@ -28,33 +29,43 @@ def annotations_to_coco(
...
+        will not include the file extension as these frame files
+        may not exist (e.g. if the frames files have not been
+        extracted from the clip).
 
     Returns
     -------
-    pathlib.Path
-        Path to the saved ``poseinterface`` COCO JSON file.
+    Path
+        Path to the saved COCO JSON file.
```
sfmig:
```suggestion
        Path to the saved ``poseinterface`` COCO JSON file. 
```

poseinterface/io.py
```diff
@@ -94,18 +103,19 @@ def annotations_to_coco(
...
-        image_filenames=coco_image_filenames,
-        visibility_encoding=coco_visibility_encoding,
+        sub_id=sub_id,
+        ses_id=ses_id,
+        cam_id=cam_id,
+        include_file_extension=(mode == "frame"),
     )
+    # Generate COCO dict
+    coco_data = coco.convert_labels(labels, image_filenames=image_filenames)
+    # Update image IDs in coco_data to match the frame IDs in the filenames
```
sfmig:
```suggestion
    # Update image IDs in coco_data to match the frame numbers from the filenames
```
Since we call them frame numbers in the rest of this file

poseinterface/io.py
```diff
@@ -28,33 +29,43 @@ def annotations_to_coco(
...
+        Path to the saved COCO JSON file.
+
+    Raises
+    ------
+    ValueError
+        If no labeled frames could be extracted from the input file,
+        or if the annotations refer to multiple videos.
 
     Notes
     -----
```
sfmig, line 70: We can link to the relevant specs in the notes.

For the frame labels output, see the spec at https://poseinterface.neuroinformatics.dev/project_structure.html#frame-labels-framelabels-json. 

For the clip labels output, see https://poseinterface.neuroinformatics.dev/project_structure.html#clip-labels-cliplabels-json.

For the start labels output, see https://poseinterface.neuroinformatics.dev/project_structure.html#clip-start-labels-startlabels-json

poseinterface/io.py
```diff
@@ -28,33 +29,43 @@ def annotations_to_coco(
...
-        but not visible, 2: labeled and visible) or "binary" (0: not
-        visible, 1: visible). Default is "ternary".
+    sub_id
+        Subject ID to include in the generated filenames.
+    ses_id
+        Session ID to include in the generated filenames.
+    cam_id
+        Camera ID to include in the generated filenames.
+    mode
+        Whether to generate framelabels.json or cliplabels.json.
```
sfmig:
```suggestion
        Whether to generate framelabels.json, cliplabels.json, or startlabels.json.
```
Right? I think the last two are the same mode though. We'd need to paraphrase a bit.

poseinterface/io.py
```diff
@@ -158,3 +168,81 @@ def _extract_frame_number(
...
+
+def _generate_poseinterface_filenames(
+    labels: sio.Labels,
+    *,
+    sub_id: str,
+    ses_id: str,
+    cam_id: str,
+    include_file_extension: bool = False,
+) -> list[str]:
+    """Generate PoseInterface image filenames an input annotations file.
```
sfmig:
```suggestion
    """Generate PoseInterface image filenames for an input annotations file.
```

poseinterface/io.py
```diff
@@ -158,3 +168,81 @@ def _extract_frame_number(
...
+
+    Raises
+    ------
+    ValueError
+        If no labeled frames could be extracted from the input file.
+    """
+    video_filenames = labels.videos[0].filename
+    if isinstance(video_filenames, list):  # Sequence of frame images
+        frame_numbers = [
+            _extract_frame_number(Path(fn).stem, frame_regexp=r"(\d+)")
```
sfmig:
```suggestion
            _extract_frame_number(Path(fn).stem, frame_regexp=INPUT_FILE_FRAME_REGEXP)
```

And then `INPUT_FILE_FRAME_REGEXP=r"(\d+)"` at the top. Just to not hide this assumption here.

poseinterface/io.py
```diff
@@ -158,3 +168,81 @@ def _extract_frame_number(
...
+        file_extensions = (
+            [Path(fn).suffix for fn in video_filenames]
+            if include_file_extension
+            else []
+        )
+    else:  # Video file
+        frame_numbers = [lf.frame_idx for lf in labels.labeled_frames]
+        file_extensions = (
+            [".png"] * len(frame_numbers) if include_file_extension else []
+        )
```
sfmig, line 299: We should say somewhere that in this case we assume the extension is `.png`

tests/data/sleap/sub-testSub123_ses-testSes123_cam-testCam123_start-1000_dur-5_cliplabels.json
```diff
@@ -0,0 +1,250 @@
+{
+    "images": [
+        {
+            "id": 1000,
+            "file_name": "sub-testSub123_ses-testSes123_cam-testCam123_frame-1000",
+            "width": 0,
+            "height": 0
```
sfmig, line 7: should we include a width, height that is consistent with the kpt coords/bbox below?

tests/test_integration/test_io.py
```diff
@@ -12,12 +12,12 @@
         "dlc_multi_index_in_project_root",
     ],
 )
-def test_annotations_to_coco(input_path, tmp_path, request):
+def test_annotations_to_coco(input_path, tmp_path, test_ids, request):
```
sfmig: To be renamed if we rename the function too.

tests/test_integration/test_io.py
```diff
@@ -12,12 +12,12 @@
...
     """Test that annotations in different project structures are converted."""
 
     input_path = request.getfixturevalue(input_path)
     output_json_path = tmp_path / "output.json"
 
-    annotations_to_coco(input_path, output_json_path)
+    annotations_to_coco(input_path, output_json_path, **test_ids)
```
sfmig:
```suggestion
    annotations_to_coco(input_path, output_json_path, **sub_ses_cam_ids)
```
A bit more explicit

tests/conftest.py
```diff
@@ -7,8 +7,19 @@
...
 from PIL import Image
 
-# Path to the directory containing the test data
 TEST_DATA_DIR = Path(__file__).parent / "data"
+TEST_DATA_DLC_DIR = TEST_DATA_DIR / "dlc"
+TEST_DATA_SLEAP_DIR = TEST_DATA_DIR / "sleap"
+
+
+@pytest.fixture
+def test_ids():
```
sfmig:
```suggestion
def sub_ses_cam_ids():
```

tests/test_unit/test_io.py
```diff
@@ -18,26 +22,32 @@ def test_annotations_to_coco(
...
     # Check subfunctions are all called
     mock_load_file.assert_called_once_with(input_csv)
     mock_convert_labels.assert_called_once_with(
         mock_labels,
-        image_filenames=None,
-        visibility_encoding="ternary",
+        image_filenames=[
+            "sub-testSub123_ses-testSes123_cam-testCam123_frame-0.png"
+        ],
     )
```
sfmig, line 73: I know they are tested separately but since this test checks the wiring within `annotations_to_coco`, I think the calls to  `_generate_poseinterface_filenames` and `_update_image_ids` should also be checked here

tests/test_integration/test_io.py
sfmig: Since we use lazy fixtures below, maybe we can replace `request` with the lazy fixture wrapper here too:

```python
from pytest_lazy_fixtures import lf

@pytest.mark.parametrize(
    "input_path",
    [
        lf("dlc_single_index_in_video_folder"),
        lf("dlc_multi_index_in_video_folder"),
        lf("dlc_single_index_in_project_root"),
        lf("dlc_multi_index_in_project_root"),
    ],
)
def test_annotations_to_coco(input_path, tmp_path, test_ids):
    ...
```

sfmig (changes requested):
Thanks @lochhh! Just a few minor comments

examples/SWC-plusmaze_to_benchmark.py
```diff
@@ -98,44 +96,18 @@
...
 # %%
 # Convert DLC annotations to COCO format
 # --------------------------------------
 # Here we use the :func:`annotations_to_coco` function from `poseinterface.io`
 # which wraps around `sleap_io` functionality to perform the conversion.
-# We pass the new filenames so the JSON references the renamed frames.
 
 annotations_to_coco(
     input_path=source_annotations_path,
     output_json_path=target_annotations_path,
```
sfmig: Shouldn't `output_json_path` be determined by the spec and the input `sub_id`, `ses_id`, `cam_id`? 

Should `output_json_path` be instead the path to the parent directory holding the `_framelabels.json` or the `_cliplabels.json` file?

poseinterface/io.py
```diff
@@ -158,3 +168,81 @@ def _extract_frame_number(
...
+    {sub_id}_{ses_id}_{cam_id}_frame-{0-padded_frame_number}
+
+    If `include_file_extension` is True, the generated filenames will include
+    the file extension of the original frame files, in the format:
+    {sub_id}_{ses_id}_{cam_id}_frame-{0-padded_frame_number}.{file_extension}
+
+    Parameters
+    ----------
+    input_path
+        Path to the input annotations file.
```
sfmig:
```suggestion
    labels
        ...
```

poseinterface/io.py
```diff
@@ -28,33 +29,43 @@ def annotations_to_coco(
...
-        If provided, must be a single string (for single-frame videos)
-        or a list of strings matching the number of labeled frames.
-        If None (default), generates filenames from video filenames
-        and frame indices.
-    coco_visibility_encoding : str, optional
-        Encoding scheme for keypoint visibility in the ``poseinterface`` COCO
-        JSON file. Options are "ternary" (0: not labeled, 1: labeled
-        but not visible, 2: labeled and visible) or "binary" (0: not
-        visible, 1: visible). Default is "ternary".
+    sub_id
```
niksirbi, line 48: While playing with using this updated function to convert data, I wondered whether we even need to expose `sub_id`, `ses_id` and `cam_id` here.

When converting data to the `poseinterface` format, the `output_json_path` will be sth like `sub-M708149_ses-20200317_cam-topdown_framelabels.json`, which already has all the information we need.

In my scripts, it made more sense to first figure out the target video filename (from which all other names follow), before I ran `annotations_to_coco`.  The current implementation here assumes that the user will have figured out the correct `sub_id`, `ses_id` and `cam_id`, but for some reason won't use them for the .json filename.

I guess my question is about how much we want to restrict the use-case here. If we only use this function to convert to our `poseinterface` format, we can assume standard naming for the output json file. I personally think that's a fair assumption, because `sub_id`, `ses_id`, and `cam_id` would anyway make no sense in other contexts.

1 directories, 2 files

poseinterface/io.py
```diff
@@ -158,3 +168,81 @@ def _extract_frame_number(
...
+        List of generated COCO image filenames corresponding to each
+        labeled frame.
+
+    Raises
+    ------
+    ValueError
+        If no labeled frames could be extracted from the input file.
+    """
+    video_filenames = labels.videos[0].filename
+    if isinstance(video_filenames, list):  # Sequence of frame images
```
niksirbi, line 285: I encountered an error here when running this on a specific dataset.

<details>
<summary>Expand to see traceback</summary>

```pytb
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[9], line 39
     35 # Convert annotations from DLC CSV to COCO JSON
     36 framelabels_path = (
     37     target_frames_dir / f"{video_prefix}_framelabels.json"
     38 )
---> 39 annotations_to_coco(
     40     input_path=(
     41         source_frames_dir / "CollectedData_Loukia.csv"
     42     ),
     43     output_json_path=framelabels_path,
     44     sub_id=subject_id,
     45     ses_id=session_id,
     46     cam_id=camera_id,
     47 )
     49 # Copy frames, renaming per the COCO JSON filenames.
     50 copy_frames_with_coco_names(
     51     source_dir=source_frames_dir,
     52     target_dir=target_frames_dir,
     53     framelabels_path=framelabels_path,
     54 )

File ~/Code/NIU/poseinterface/poseinterface/io.py:116, in annotations_to_coco(input_path, output_json_path, sub_id, ses_id, cam_id, mode)
    108 image_filenames = _generate_poseinterface_filenames(
    109     labels,
    110     sub_id=sub_id,
   (...)    113     include_file_extension=(mode == "frame"),
    114 )
    115 # Generate COCO dict
--> 116 coco_data = coco.convert_labels(labels, image_filenames=image_filenames)
    117 # Update image IDs in coco_data to match the frame IDs in the filenames
    118 coco_data = _update_image_ids(coco_data)

File ~/Code/NIU/poseinterface/.venv/lib/python3.13/site-packages/sleap_io/io/coco.py:496, in convert_labels(labels, image_filenames, visibility_encoding)
    494         image_filenames = [image_filenames]
    495     if len(image_filenames) != len(labels.labeled_frames):
--> 496         raise ValueError(
    497             f"Number of image filenames ({len(image_filenames)}) must match "
    498             f"number of labeled frames ({len(labels.labeled_frames)})"
    499         )
    501 # Process labeled frames
    502 image_id_counter = 1

ValueError: Number of image filenames (16) must match number of labeled frames (15)
```
</details>

Claude suggested this fix, which worked (at least for the dataset I was working on).

```python
    if isinstance(video_filenames, list):  # Sequence of frame images
        # Use labeled_frames to select only frames with annotations,
        # since video_filenames may include unlabelled images.
        labeled_filenames = [
            video_filenames[lf.frame_idx]
            for lf in labels.labeled_frames
        ]
        frame_numbers = [
            _extract_frame_number(Path(fn).stem, frame_regexp=r"(\d+)")
            for fn in labeled_filenames
        ]
        file_extensions = (
            [Path(fn).suffix for fn in labeled_filenames]
            if include_file_extension
            else []
        )
```

niksirbi: I encountered the above error on the following session (on `neuroinformatics` ceph partition):

```
/ceph/neuroinformatics/neuroinformatics/sirmpilatzen/behav_data/Loukia/MASTER_DoNotModify/MouseTopDown-Loukia-2022-09-13/labeled-data/M723400_EPM_20200317_152959980-converted
```

niksirbi: I suspect this may be related to the fact that some sampled frames for this session contain no annotations (the animal is totally missing).

poseinterface/io.py
lochhh: I used [Sphinx cross references here](https://github.com/neuroinformatics-unit/poseinterface/pull/26/changes/87e5d1b20a8de83619a7b01658ef52151b03165a), but it won't be very helpful outside of the website... On the other hand using the fixed URLs means we need to always remember to update them.

poseinterface/io.py
lochhh: "Generate PoseInterface image filenames for frames in the input labels." beacuse the input is a SLEAP labels object

poseinterface/io.py
lochhh: I went with [DLC_FRAME_REGEXP](https://github.com/neuroinformatics-unit/poseinterface/pull/26/changes/9e6337aae551c024cd972544adcb83f9ed5d32be) because this is the assumption in sleap io

poseinterface/io.py
lochhh: I [added a notes section](https://github.com/neuroinformatics-unit/poseinterface/pull/26/changes/87e5d1b20a8de83619a7b01658ef52151b03165a) for this internal function, but this won't be visible from annotations_to_poseinterface

tests/data/sleap/sub-testSub123_ses-testSes123_cam-testCam123_start-1000_dur-5_cliplabels.json
lochhh: sleap-io auto-generates this based on the video file, but because this video does not exist, the video shape is None. In the other generated framelabels.json the width and height are 1px because we generated fake 1x1 images. I would leave it as is, since we plan to use "real" test data in the near future (perhaps the one @niksirbi just put on [Dropbox](https://www.dropbox.com/scl/fo/zxwqnhx8e413xofaaa5iv/AK6LL6iTdfVz1oT_jub5IJA?dl=0&rlkey=xxl3z1du4jcl4aa0c6a8r1et9)).

tests/test_integration/test_io.py
lochhh: Done: [Extend integration test to all formats and use lazyfixtures](https://github.com/neuroinformatics-unit/poseinterface/pull/26/changes/970478550abd3ca8e896d981a231c44e5a9c4fae)

tests/test_integration/test_io.py
lochhh: [Rename test_ids fixture to sub_ses_cam_ids](https://github.com/neuroinformatics-unit/poseinterface/pull/26/changes/aa291b6934f5041da2c035a19c86c5e350285af6)

tests/test_unit/test_io.py
lochhh: I've [refactored the test to test its behaviour](https://github.com/neuroinformatics-unit/poseinterface/pull/26/changes/651085a65e4acf0bf64b5df36014b71dbdae4e1f#diff-e404e6ee8030473de636237b9c3277ca0d2e60db76c49aa361ad7721c6ce18f2), i.e. that the expected file is generated with the expected content and at the expected location, while mocking only external function returns. This indirectly checks each subfunction is called.

poseinterface/io.py
lochhh: Opened #44

tests/data/sleap/sub-testSub123_ses-testSes123_cam-testCam123_start-1000_dur-5_cliplabels.json
niksirbi: Agreed with CH, with the caveat that the Dropbox data is not ready-to-use yet. I need to double-check it and fix a few issues first.

poseinterface/io.py
```diff
@@ -94,28 +109,69 @@ def annotations_to_coco(
...
+        )
+    start_frame = min(image_ids)
+    n_frames = len(image_ids)
+    return (
+        output_dir
+        / f"{prefix}_start-{start_frame}_dur-{n_frames}_{label_suffix}.json"
+    )
+
+
+def _update_image_ids(coco_data: dict) -> dict:
```
niksirbi: `_update_image_ids` always sets each image id to the session-video frame number extracted from the filename, which is correct for framelabels but violates the spec for clip and start formats. Per the [spec](https://poseinterface.neuroinformatics.dev/benchmark-dataset.html), clip image IDs must be 0-based within the clip (0, 1, 2, ...). The current test cliplabels JSON confirms the bug; IDs are 1000-1004 where they should be 0-4. The fix would be to enumerate rather than extract from the filename when format != "frame".

niksirbi: Btw `_build_output_json_path` will break after fixing this bug: once cliplabels IDs become 0-based within the clip, `min(image_ids)` will always return 0 instead of the actual session frame index, breaking the start frame calculation. So this has to be adjusted as well.

poseinterface/io.py
```diff
@@ -94,28 +109,69 @@ def annotations_to_coco(
...
+    if len(image_ids) == 0:
+        raise ValueError(
+            "No image IDs were found in the COCO data. "
+            f"Cannot infer start frame and duration for {label_suffix} format."
+        )
+    start_frame = min(image_ids)
+    n_frames = len(image_ids)
+    return (
+        output_dir
+        / f"{prefix}_start-{start_frame}_dur-{n_frames}_{label_suffix}.json"
```
niksirbi: The spec says:

> <frameID> in the start field must be the 0-based index of the first frame of the clip in the session video, padded to a consistent width (e.g. 0500, 1000).

There is no padding performed here. We should either change the code (and corresponding tests) to zero-pad based the largest frame index, or we should relax the spec in that regard.

poseinterface/io.py
```diff
@@ -2,11 +2,14 @@
 import json
 import re
 from pathlib import Path
+from typing import Literal, TypeAlias
 
 import sleap_io as sio
 from sleap_io.io import coco
 from sleap_io.io.dlc import is_dlc_file
 
+PoseInterfaceFormat: TypeAlias = Literal["clip", "frame", "start"]
```
niksirbi: Not sure we need `start` as an option. In practice, `startlabels` will be auto-derived from `cliplabels` by just keeping the first entry.
The current code allows users to derived a file suffixed with `_startlabels` but with multiple entries, which is essentially `cliplabels`.
I think it's cleanest to just do away with `start` as an option. Instead, we need a standalone utility that converts `cliplabels` to `startlabels` by renaming the suffix and keeping just the first entry (probably in a separate future PR).

niksirbi (changes requested):
I've gone through the PR one additional time, using everything I've learned while working on #40.

**The good news**: I think we can consider all previous review comments as addressed.
- We now have a separate issue #44 to track (and a corresponding PR upstream in `sleap-io`).
- Regarding the `sub/ses/cam_id` kwargs, we should leave it as is for now. We can revisit function signatures after we've merged all the individual functions (`annotation_to_poseinterface`, `predictions_to_poseinterface`, `video_to_poseinterface`, etc.). We can then come up with a consistent scheme that makes sense.

**The bad news**: I found some new bugs during this review.

There are 3 substantive issues related to discrepancies between this PR and the [spec](https://poseinterface.neuroinformatics.dev/project_structure.html).
These are all intertwined with each other:

1. Image IDs in cliplabels do not follow the spec. `_update_image_ids` always sets each image `id` to the session-video frame number extracted from the filename. For `framelabels` that is correct, but the [spec requires](https://poseinterface.neuroinformatics.dev/project_structure.html#clip-labels-cliplabels-json) clip image IDs to be 0-based within the clip (`0, 1, 2, ...`). The committed test cliplabels JSON confirms the bug, IDs are `1000–1004` where they should be `0–4`.
2. The `start` field in the output filename is not padded. `_build_output_json_path` writes `start-{start_frame}` as a bare integer. The spec requires it to be padded to a consistent width. We should either pad or relax the spec. The integration test hardcodes the unpadded values and would need updating too.
3. Fixing (1) breaks the start-frame derivation in `_build_output_json_path`. The start frame is currently computed as `min(image_ids)`. That only works because image IDs are session-video indices in the current implementation (issue 1). Once IDs are corrected to `0, 1, 2, ...`, `min(image_ids)` will always return `0`.

We may be able to fix all 3 in one fell swoop: extract the start frame from `file_name` first (resolving 3), use that value (padded to the same width as the frame IDs) in the filename (resolving 2), then re-index image IDs to `0, 1, 2, ...` separately (resolving 1).

Other than these I also came across a minor testing bug: a swap in the order of `@patch` pytest decorators (they are applied in the inverse order than one would expect). I've pushed a fix for that in [76b7767](https://github.com/neuroinformatics-unit/poseinterface/pull/26/commits/76b7767d16c9b300020cc8bb5cbfca3123fbe639).

niksirbi:
@lochhh I've created PR #48 with my suggested fixes to the aforementioned issues. Feel free to grab from that whatever you wish.

poseinterface/io.py
lochhh: I initially thought it would be nice to be able to generate startlabels directly, without needing to first generate cliplabels, but I guess we will always have a corresponding cliplabels file for every startlabels file we create, so your proposed idea makes a lot of sense.

niksirbi (approved):
LGMT now 🚀

Issue #10, opened 2025-12-17
Utility for extracting a clip from a video


niksirbi:
Given an start frame and a duration.

@sfmig knows how to do this with `ffmpeg`.
Maybe we can try using sleap-io for this, and test against Sofía's existing ffmpeg approach.

niksirbi:
Initially we though of this utility as: given 1 video + start and duration (in frames), return a clip from that video.

In discussions with @sfmig today we thought that maybe this function could also optionally accept a `cliplabels.json` file corresponding to the entire video, and return a smaller `cliplabels.json` file for the same frames as the extracted clip.

So:
- Minimal case: `video` -> `clip` (subset of video)
- Nice to have: `video` + `cliplabels.json` -> `clip` (subset of video) + smaller `cliplabels.json` (subset of input `cliplabels.json`)

The input `cliplabels.json` (the one for the whole video) will likely come from converting a file containing keypoint predictions, see #38.

niksirbi:
The implementation in #39 follows the 2nd option from above, i.e.:
`video` + `cliplabels.json` -> `clip` (subset of video) + smaller `cliplabels.json` (subset of input `cliplabels.json`)
- The input `cliplabels.json` holds the predictions corresponding to the entire session video, converted to our [cliplabels.json](https://poseinterface.neuroinformatics.dev/benchmark-dataset.html#clip-labels-cliplabels-json) flavour of the COCO `.json` format by #38.
- The output `cliplabels.json` holds the same labels, but only for the corresponding extracted video clip. These files will be manually proof-read and corrected by experts to generate the ground-truth labels for the benchmark.

We reasoned that conversion from source format (DLC, SLEAP, etc.) to poseinterface format would occur only once, whereas clip extraction may occur several times (clips may be extracted at various time points after the conversion, as needed). In such a workflow, it makes sense to cache the predicted labels corresponding to the entire video, so we can avoid repeating the `predictions_to_poseinterface` step. See the image representing the envisioned workflow, using a DLC project as the source:

<img width="2041" height="863" alt="Image" src="https://github.com/user-attachments/assets/1738678d-cf7d-4c43-bd6b-83fed22d317d" />

Several questions arise from this choice:
[...]

PR #39, opened 2026-03-27
Format videos to `poseinterface` spec. Extract clip function. 

sfmig:
## Description

**What is this PR**

- [ ] Bug fix
- [ ] Other

**What does this PR do?**
- Added `video_to_poseinterface` to io module, to convert a video to `poseinterface` format.
	- It renames videos following spec and reencodes them if required. 
- Basic util to extract a clip and the corresponding `cliplabels.json` file, given a video in `poseinterface` format, its full video annotations in `cliplabels.json` format and a range of frames.
	- Exposed as entry point `extract-clip` 	

## References

\

Tests pass locally and in CI.

## Is this a breaking change?

No.

Yes, docstrings.

## Checklist:


poseinterface/clips.py
```diff
@@ -0,0 +1,185 @@
...
+    if start_frame + duration > video.shape[0]:
+        duration = video.shape[0] - start_frame
+        logging.warning(
+            "Clip exceeds video length. "
+            f"Clamping duration to {duration} frames."
+        )
+
+    # Slice clip and save as mp4
+    clip = video[start_frame : start_frame + duration]
+    clip_path = (
```
niksirbi, line 98: This should be `video_path.stem` not `video.stem`, right?

```
    clip_path = (
        clips_dir / f"{video_path.stem}_start-{start_frame}_dur-{duration}.mp4"
    )
```

sfmig: both are equivalent, `video.stem` gets the filename from the sleap-io Video object

poseinterface/io.py
```diff
@@ -23,6 +28,19 @@
...
+# We support sleap's MediaVideo files
+EXPECTED_SUFFIX = ".mp4"
+EXPECTED_ENCODING = {
+    "pixelformat": "yuv420p",
+    "codec": "h264",  # codec name
+}
+REENCODING_PARAMS = {
+    **EXPECTED_ENCODING,
+    "codec": "libx264",  # overwrite with encoder to use
+    "crf": 25,
```
niksirbi, line 43: SLEAP's (and therefore OCTRON's) magic incantation uses crf 23. Any specific reason for going with 25 here?

poseinterface/io.py
```diff
@@ -1,10 +1,15 @@
+"""Functions to convert annotations and videos to PoseInterface format."""
```
niksirbi: Everywhere else we style package's name as lowercase, usually monospace a la `movement`. I recommend we don't go into CamelCase, unless we make an explicity decision to do so project-wide.

```suggestion
"""Functions to convert annotations and videos to ``poseinterface`` format."""
```

poseinterface/io.py
```diff
@@ -158,3 +176,135 @@ def _extract_frame_number(
...
+    # Check if suffix is mp4
+    if input_video_path.suffix.lower() != EXPECTED_SUFFIX:
+        return True
+
+    # Check codec and pixelformat
+    encoding = _get_codec_pixelformat(input_video_path)
+    if encoding != EXPECTED_ENCODING:
+        logging.warning(
+            f"Video encoding {encoding} does not match "
+            f"expected {EXPECTED_ENCODING}. Please reencode "
```
niksirbi: Since re-encoding happens automatically if needed. Should we reframe this as 'Will reencode' instead of 'Please reencode'?

Also since this is actually an expected action (documented in the docstring of `video_to_poseinterface`), I wonder whether this should be an INFO instead of WARNING.

poseinterface/io.py
niksirbi: Releatedly, I wonder whether we should expose crf to the user as an optional kwarg in `video_to_poseinterface`

poseinterface/clips.py
```diff
@@ -0,0 +1,185 @@
+"""Functions to extract clips from poseinterface videos."""
```
niksirbi:
```suggestion
"""Functions to extract clips from ``poseinterface`` videos."""
```

poseinterface/clips.py
```diff
@@ -0,0 +1,185 @@
...
+
+    # Generate cliplabels.json from the full video labels
+    clip_json = _extract_cliplabels(
+        video_path, clips_dir, start_frame, duration
+    )
+
+    return clip_path, clip_json
+
+
+def _extract_cliplabels(video_path, clips_dir, start_frame, duration):
```
niksirbi: Quoting our spec on `cliplabels.json`:

> * Clip labels follow the same COCO keypoints format as frame labels, but with different conventions for image `id` and `file_name` values:
> * Each image `id` *must* be the **0-based index of the frame within the clip** (i.e. `0`, `1`, `2`, ...), not the index in the session video.
> * Each `file_name` *must* follow the same pattern as [frame image filenames](#frames), but **without the extension**. The `frame` field in the `file_name` *must* correspond to the index of that frame in the **session video**.
> 
> This means that each entry in the `images` array encodes two pieces of information: the `id` gives the local position within the clip, while the `frame` field in `file_name` gives the global position in the session video. Note that in both cases the indices are 0-based.
> 
> For a clip starting at frame 1000 with a duration of 5 frames, the `images` array would be:
> 
> ```json
> [
> {"id": 0, "file_name": "sub-M708149_ses-20200317_cam-topdown_frame-1000", "width": 1300, "height": 1028},
> {"id": 1, "file_name": "sub-M708149_ses-20200317_cam-topdown_frame-1001", "width": 1300, "height": 1028},
> {"id": 2, "file_name": "sub-M708149_ses-20200317_cam-topdown_frame-1002", "width": 1300, "height": 1028},
> {"id": 3, "file_name": "sub-M708149_ses-20200317_cam-topdown_frame-1003", "width": 1300, "height": 1028},
> {"id": 4, "file_name": "sub-M708149_ses-20200317_cam-topdown_frame-1004", "width": 1300, "height": 1028}
> ]
> ```

niksirbi: This function correctly selects the images corresponding to clip, but there is a step missing: the IDs of the extracted images must be changed to start with 0 within the extracted clip (i.e. subtract `start_frame`). The `file_name`s should be left as they are, to keep a reference to the global frame index.

poseinterface/clips.py
niksirbi: Not sure these are exactly equivalent. While trying this PR on some real data, I encountered the following error:

```pytb
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[10], line 28
     25 session_dir = benchmark_base_dir / split / project_name / sub_ses_prefix
     27 for start_frame in start_frames:
---> 28     clip_path, clip_json = extract_clip(
     29         video_path=(session_dir / f"{sub_ses_cam_prefix}.mp4"),
     30         start_frame=start_frame,
     31         duration=duration,
     32     )
     33     print(f"Extracted clip: {clip_path}, {clip_json}")

File ~/Code/NIU/poseinterface/poseinterface/clips.py:85, in extract_clip(video_path, start_frame, duration)
     82 # Slice clip and save as mp4
     83 clip = video[start_frame : start_frame + duration]
     84 clip_path = (
---> 85     clips_dir / f"{video.stem}_start-{start_frame}_dur-{duration}.mp4"
     86 )
     87 sio.save_video(clip, clip_path, fps=video.fps)
     89 # Generate cliplabels.json from the full video labels

AttributeError: 'Video' object has no attribute 'stem'
 ```
 
 The error goes away when using my version of `clip_path` (from my previous comment).

File: poseinterface/clips.py (hunk 6 above)
Comment (niksirbi) [reply poseinterface/clips.py]: A similar problem applies to the annotation ids inside the extracted cliplabels file. For a clip starting at frame 1000, the first annotation entry returned by the current implementation is as follows:

```json
 "annotations": [
    {
      "id": 1001,
      "image_id": 1000,
      "category_id": 1,
      "keypoints": [529.621887207031, 494.971038818359, 2, 543.039184570313, 501.402648925781, 2, 544.258728027344, 482.781982421875, 2, 599.23681640625, 496.673095703125, 2, 593.133361816406, 527.087524414063, 2, 604.429321289063, 470.6630859375, 2, 669.785888671875, 507.377227783203, 2, 673.556396484375, 613.862365722656, 2],
      "num_keypoints": 8,
      "bbox": [529.621887207031, 470.6630859375, 143.934509277344, 143.199279785156],
      "area": 20611.3180647455,
      "iscrowd": 0
    },
```

Since our spec recommends that annotation IDs are 1-indexed, the "id" here should probably be 1, with the "image_id" being 0, as per my previous comment.

poseinterface/clips.py
```diff
@@ -0,0 +1,185 @@
...
+    clip = video[start_frame : start_frame + duration]
+    clip_path = (
+        clips_dir / f"{video.stem}_start-{start_frame}_dur-{duration}.mp4"
+    )
+    sio.save_video(clip, clip_path, fps=video.fps)
+
+    # Generate cliplabels.json from the full video labels
+    clip_json = _extract_cliplabels(
+        video_path, clips_dir, start_frame, duration
+    )
```
niksirbi: I think this step should be optional, as in, only extract cliplabels if a corresponding (appropriately named) .json file is to be found in the same folder as the video, otherwise just do the video clip. This will make `extract_clip` broadly useful in all sorts of contexts (beyond the specific purpose of generating clips for poseinterface benchmarks).

As things are, you can't really use `extract_clip`, unless you have the companion .json file.

poseinterface/clips.py
```diff
@@ -0,0 +1,185 @@
...
+
+    return clip_path, clip_json
+
+
+def _extract_cliplabels(video_path, clips_dir, start_frame, duration):
+    """Extract clip labels from the video cliplabels.json file."""
+    # Read file with labels for the whole video
+    video_json = video_path.parent / f"{video_path.stem}_cliplabels.json"
```
niksirbi: I'm no longer certain about the suffix of this json file, see this comment https://github.com/neuroinformatics-unit/poseinterface/issues/10#issuecomment-4281509141 and the discussion in the PR review for #45.

poseinterface/clips.py
```diff
@@ -0,0 +1,185 @@
...
+        clips_dir / f"{video.stem}_start-{start_frame}_dur-{duration}.mp4"
+    )
+    sio.save_video(clip, clip_path, fps=video.fps)
+
+    # Generate cliplabels.json from the full video labels
+    clip_json = _extract_cliplabels(
+        video_path, clips_dir, start_frame, duration
+    )
+
+    return clip_path, clip_json
```
niksirbi, line 121: Would be nice to log an INFO message about this function's success before returning, to signal that it has actually worked.

niksirbi (changes requested):
Thanks @sfmig.

I tried using the new functions in #40. They mostly worked, but I've stumbled on a few issues that need to be addressed before merging (see inline comments). Happy to do another round of review after we resolve these (I skipped the tests in this round).

By the way, if we want the new public functions to appear in the API references, we have to add them manually in `api_index.rst` (we haven't yet set up the automatic machinery we have in `movement`).

poseinterface/io.py
sfmig: I used `crf=25` because it is the [default value used in sleap-io](https://github.com/talmolab/sleap-io/blob/354419a4bd476db9d3819527524b901d88995467/sleap_io/io/video_writing.py#L57). Seems a more reliable/tractable source than a command in the docs, although I expect the difference to be minimal.

I think for now it makes sense to keep the encoding parameters as hardcoded defaults (with the REENCODING_PARAMS). It makes it simpler for the users and reflects our intention of unifying the video inputs.

poseinterface/clips.py
sfmig: apologies, I think I checked quickly at the time and mixed up the two variables

poseinterface/clips.py
sfmig: Thanks for the feedback - please see [this comment](https://github.com/neuroinformatics-unit/poseinterface/issues/10#issuecomment-4430228528) for the changes I implemented re this point

sfmig:
Hi @niksirbi , thanks for the review and apologies for the delay. I believe this is now ready to go, will request a re-review now.

niksirbi (approved):
Thanks for updating @sfmig!

I'm happy with it as is. I've merged #26, then rebased this one to `main` (was painful) and pushed some very minor cosmetic changes after that. 

It's good to go now!

PR #45, opened 2026-04-01
Convert predictions to cliplabels.json using movement

### LINKED ISSUE #38: Implement a `predictions_to_coco` function
We currently have `annotations_to_coco`, but we also need to convert prediction files to the COCO .json format, to facilitate the extraction of `cliplabels.json` (see #10).

For SLEAP .slp file, it could be that `annotation_to_coco` can already do this? Though we'd need to double-check how labeled vs predicted instances are handled by `sleap-io` here.

For DLC (and LP) files, I suspect we'll for sure need a separate `predictions_to_coco` function. We'd discussed maybe doing it via `movement` (e.g. DLC predictions -> `movement` dataset -> SLEAP file -> `sleap-io` -> COCO), but:
- Not sure that would work
- Probably it's more efficient to not go via `movement` in this case.

> [!note]
> Because we are talking about predictions inputs here, we can probably use `movement` sample data for testing.

sfmig:
## Description

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**
To convert prediction files supported by `movement` to `cliplabels.json` files, that we can clip later when extracting video clips.

**What does this PR do?**
It adds a `predictions_to_poseinterface` function that reads prediction files as movement datasets and exports the trajectories as a `cliplabels.json` file, defined for the full range of frames in the prediction file.

## References

\

Tests pass locally and in CI.

## Is this a breaking change?
No.

No.

## Checklist:


poseinterface/io.py
```diff
@@ -158,3 +195,224 @@ def _extract_frame_number(
...
+        ses_id=ses_id,
+        cam_id=cam_id,
+        img_h=img_h,
+        img_w=img_w,
+    )
+
+    # Export dict as JSON
+    output_json_parent_dir = (
+        Path(output_json_parent_dir)
+        / f"sub-{sub_id}_ses-{ses_id}_cam-{cam_id}.json"
```
lochhh: Based on @niksirbi 's excalidraw illustration
```suggestion
        / f"sub-{sub_id}_ses-{ses_id}_cam-{cam_id}_cliplabels.json"
```

poseinterface/io.py
```diff
@@ -158,3 +195,224 @@ def _extract_frame_number(
...
+        ds,
+        sub_id=sub_id,
+        ses_id=ses_id,
+        cam_id=cam_id,
+        img_h=img_h,
+        img_w=img_w,
+    )
+
+    # Export dict as JSON
+    output_json_parent_dir = (
```
lochhh:
```suggestion
    output_json_path = (
```

poseinterface/io.py
```diff
@@ -158,3 +195,224 @@ def _extract_frame_number(
...
+        Path(output_json_parent_dir)
+        / f"sub-{sub_id}_ses-{ses_id}_cam-{cam_id}.json"
+    )
+    with open(output_json_parent_dir, "w") as f:
```
lochhh:
```suggestion
    with open(output_json_path, "w") as f:
```

poseinterface/io.py
```diff
@@ -158,3 +195,224 @@ def _extract_frame_number(
...
+        json.dump(coco_data, f)
+
+    return output_json_parent_dir
```
lochhh:
```suggestion
    return output_json_path
```

poseinterface/io.py
```diff
@@ -158,3 +195,224 @@ def _extract_frame_number(
...
+    # with models that treat category 0 as background.
+    categories = [
+        {
+            "id": i + 1,
+            "name": name,
+            "keypoints": keypoint_names,
+            "skeleton": [],
+        }
+        for i, name in enumerate(individual_names)
+    ]
```
lochhh, line 592:
```suggestion
    categories = [
        {
            "id": i,
            "name": name,
            "keypoints": keypoint_names,
            "skeleton": [],
        }
        for i, name in enumerate(individual_names, start=1)
    ]
```

poseinterface/io.py
```diff
@@ -158,3 +195,224 @@ def _extract_frame_number(
...
+        {
+            "id": t,
+            "file_name": (
+                f"sub-{sub_id}_ses-{ses_id}_cam-{cam_id}_frame-{t:04d}"
+            ),
+            "width": img_w,
+            "height": img_h,
+        }
+        for t in range(n_frames)
+    ]
```
lochhh, line 607: if we want dynamic width
```suggestion
    frame_idx_width = len(str(n_frames - 1))
    images = [
        {
            "id": t,
            "file_name": (
                f"sub-{sub_id}_ses-{ses_id}_cam-{cam_id}_frame-{t:0{frame_idx_width}d}"
            ),
            "width": img_w,
            "height": img_h,
        }
        for t in range(n_frames)
    ]
```

poseinterface/io.py
```diff
@@ -158,3 +195,224 @@ def _extract_frame_number(
...
+        Path(output_json_parent_dir)
+        / f"sub-{sub_id}_ses-{ses_id}_cam-{cam_id}.json"
+    )
+    with open(output_json_parent_dir, "w") as f:
+        json.dump(coco_data, f)
+
+    return output_json_parent_dir
+
+
+def _guess_source_software(file: Path | str) -> list[SourceSoftware]:
```
lochhh: this function is not covered by tests

poseinterface/io.py
```diff
@@ -158,3 +195,224 @@ def _extract_frame_number(
...
+            # dataset doesn't carry occlusion information
+            visible_array = ~np.isnan(xy[0]) & ~np.isnan(xy[1])
+            n_visible = int(visible_array.sum())
+
+            # Get list of flattened keypoints
+            # [x1, y1, v1, x2, y2, v2, ...]
+            x = np.where(visible_array, xy[0], 0.0)
+            y = np.where(visible_array, xy[1], 0.0)
+            v = np.where(visible_array, 2, 0)
+            list_xyv_kpts = np.stack([x, y, v], axis=1).ravel().tolist()
```
lochhh: can be replaced by [sleap_io.io.coco.encode_keypoints](https://io.sleap.ai/latest/reference/sleap_io/io/coco/#sleap_io.io.coco.encode_keypoints)
```suggestion
```

poseinterface/io.py
```diff
@@ -158,3 +195,224 @@ def _extract_frame_number(
...
+            else:
+                x_min, y_min, bbox_w, bbox_h = 0.0, 0.0, 0.0, 0.0
+
+            # Append results to list of annotations
+            annotations.append(
+                {
+                    "id": annot_id,
+                    "image_id": t,
+                    "category_id": i + 1,
+                    "keypoints": list_xyv_kpts,
```
lochhh:
```suggestion
                    "keypoints": coco.encode_keypoints(xy.T),
```

poseinterface/io.py
```diff
@@ -158,3 +195,224 @@ def _extract_frame_number(
...
+
+def predictions_to_poseinterface(
+    predictions_path: Path | str,
+    video_path: Path | str,
+    output_json_parent_dir: Path | str,
+    *,
+    sub_id: str,
+    ses_id: str,
+    cam_id: str,
+) -> Path:
```
lochhh, line 477: In PR #26 i'm using `input_path` and `output_dir`, wdyt about using the same here?
https://github.com/neuroinformatics-unit/poseinterface/blob/07a72900010b759ce459faa97b75eac1282f9dc4/poseinterface/io.py#L30-L38

poseinterface/io.py
```diff
@@ -158,3 +195,224 @@ def _extract_frame_number(
...
+    """Convert a prediction file to ``poseinterface`` COCO JSON format.
+
+    Reads a predictions file and writes a JSON with ``poseinterface``-style
+    filenames suitable for clip-level labels (``_cliplabels.json``).
```
lochhh: It isn't immediately clear that this function generates the "intermediary" cliplabels.json for the *session* video, and not for the *clips*. Maybe something like
```suggestion
    """Convert predictions file for a given video to ``poseinterface`` format.

    This function reads predictions for a given session video and writes the
    corresponding "session-level" COCO JSON clip labels in the
    ``poseinterface`` format, i.e.
    ``sub-<sub_id>_ses-<ses_id>_cam-<cam_id>_cliplabels.json``.

    The output JSON file is intended to be used for extracting "clip-level"
    labels, i.e.
    ``sub-<sub_id>_ses-<ses_id>_cam-<cam_id>_start-<frame_id>_dur-<n_frames>_cliplabels.json``.
```

poseinterface/io.py
```diff
@@ -158,3 +195,224 @@ def _extract_frame_number(
...
+            continue
+
+        # If suffix is covered by these validators, use them to
+        # validate the input file
+        try:
+            _validate_file(file, map_suffix_to_validators, source_software)
+            matches.append(source_software)
+        except Exception:
+            continue
+
```
lochhh: should we already raise an error when we're unable to guess the source software? otherwise taking the first guess in `predictions_to_poseinterface` would throw an IndexError when the matches list is empty. Also remember to test this behaviour
```suggestion
    if not matches:
        raise ValueError(
            f"Could not infer source software from predictions file {file}."
        )

```

poseinterface/io.py
```diff
@@ -158,3 +195,224 @@ def _extract_frame_number(
...
+    Returns
+    -------
+    list[SourceSoftware]
+        List of source software names whose validators matched.
+
+    Examples
+    --------
+    >>> from movement.io.load import guess_source_software
+    >>> guess_source_software("path/to/predictions.h5")
+    ['DeepLabCut']
```
lochhh: We can't do this in movement (yet) 😜

poseinterface/io.py
```diff
@@ -158,3 +195,224 @@ def _extract_frame_number(
...
+
+    Reads a predictions file and writes a JSON with ``poseinterface``-style
+    filenames suitable for clip-level labels (``_cliplabels.json``).
+
+    Parameters
+    ----------
+    predictions_path
+        Path to the DLC predictions CSV file.
+    video_path
+        Path to the corresponding video file.  Used to attach video
```
lochhh, line 495:
```suggestion
        Path to the corresponding session video.  Used to attach video
```

poseinterface/io.py
```diff
@@ -158,3 +195,224 @@ def _extract_frame_number(
...
+) -> Path:
+    """Convert a prediction file to ``poseinterface`` COCO JSON format.
+
+    Reads a predictions file and writes a JSON with ``poseinterface``-style
+    filenames suitable for clip-level labels (``_cliplabels.json``).
+
+    Parameters
+    ----------
+    predictions_path
+        Path to the DLC predictions CSV file.
```
lochhh: Should this be prediction file of one of the supported SourceSoftware?

tests/test_unit/test_io.py
```diff
@@ -1,17 +1,91 @@
...
+    """Mock Video object with 10 frames, matching video_labels fixture."""
+
+    def _mock_video(n_frames):
+        video = MagicMock()
+        video.fps = 30
+        video.shape = (n_frames, 480, 640, 3)
+        video.stem = "sub-01_ses-01_cam-01"
+        return video
+
+    return _mock_video
```
lochhh: This fixture is called with `mock_video(n_frames=3)` in both cases it's used. Perhaps it's sufficient to just use a SimpleNamespace, since its purpose is just for carrying attributes. Also the docstring mentions a non-existent `video_labels` fixture.
```suggestion
from types import SimpleNamespace

@pytest.fixture
def mock_video():
    """Mock Video object with 3 frames."""

    return SimpleNamespace(
        fps=30, shape=(3, 480, 640, 3), stem="sub-01_ses-01_cam-01"
    )
```

poseinterface/io.py
```diff
@@ -158,3 +195,224 @@ def _extract_frame_number(
...
+    # (take first guess)
+    source_software = _guess_source_software(predictions_path)[0]
+
+    # Read input file as movement dataset
+    # NOTE: fps=None is ignore with NWB files
+    ds = load_dataset(
+        file=predictions_path,
+        source_software=source_software,
+        fps=None,
+    )
```
lochhh: Given that _guess_source_software returns a list of candidate source software, wdyt about trying with all of these? I suppose we would succeed the first time in most cases, but could still be nice to have a fallback
```suggestion
    # Guess source software using movement validators
    source_software_guesses = _guess_source_software(predictions_path)

    # Read input file as movement dataset.
    # NOTE: fps=None is ignored with NWB files.
    ds: xr.Dataset | None = None
    load_errors: dict[SourceSoftware, str] = {}
    for source_software in source_software_guesses:
        try:
            ds = load_dataset(
                file=predictions_path,
                source_software=source_software,
                fps=None,
            )
            break
        except Exception as exc:
            load_errors[source_software] = str(exc)

    if ds is None:
        formatted_errors = "; ".join(
            f"{software}: {error}" for software, error in load_errors.items()
        )
        raise ValueError(
            "Failed to load predictions dataset with all inferred source "
            f"softwares {source_software_guesses}. Errors: "
            f"{formatted_errors}"
        )
```

tests/test_unit/test_io.py
```diff
@@ -203,3 +277,126 @@ def test_extract_frame_number_invalid(filename, frame_regexp):
...
+):
+    """Test that the relevant subfunctions are called."""
+    # Get movement dataset and video fixtures
+    ds = sample_movement_ds
+    video = mock_video(n_frames=3)
+
+    # Mock return values for supporting functions
+    mock_guess_source_software.return_value = ["DeepLabCut"]
+    mock_load_dataset.return_value = ds
+    mock_load_video.return_value.shape = video.shape
```
lochhh: mock_video has .shape, so we only patch the return value of load_video
```suggestion
    mock_load_video.return_value = video
```

poseinterface/io.py
```diff
@@ -158,3 +195,224 @@ def _extract_frame_number(
...
+
+    # Get video image width and height
+    video = sio.load_video(video_path)
```
lochhh, line 535:
```python
vid = sio.load_video("doesntexist.mp4") # This doesn't result in an error
print(vid) # Video(filename="doesntexist.mp4", shape=None, backend=MediaVideo)
_, img_h, img_w, _ = vid.shape  # TypeError: cannot unpack non-iterable NoneType object
```
We'd need extra error handling here, and make sure we cover valid and invalid cases in tests

tests/test_unit/test_io.py
```diff
@@ -203,3 +277,126 @@ def test_extract_frame_number_invalid(filename, frame_regexp):
...
+def test_predictions_to_poseinterface(
+    mock_guess_source_software,
+    mock_load_dataset,
+    mock_load_video,
+    mock_convert,
+    sample_movement_ds,
+    mock_video,
+    tmp_path,
+):
+    """Test that the relevant subfunctions are called."""
```
lochhh: I think it would be more useful to test the following behaviours rather than simply checking subfunctions are called:
- when _guess_source_software returns 0, 1, or more matches
- valid/invalid video_path (e.g. video.shape is None) 
- valid/invalid output_dir (e.g. output dir doesn't exist)
- returned file path is as expected and exists
- file content is as expected (a rough check would suffice, since the unit test for _convert_movement_ds_to_cliplabels should check this thoroughly)

We should also have basic integration tests to cover different prediction files from different source software (at least dlc and sleap?) which will test the full behaviour without mocking.

tests/test_unit/test_io.py
```diff
@@ -203,3 +277,126 @@ def test_extract_frame_number_invalid(filename, frame_regexp):
...
+    }
+
+    # Convert predictions
+    result = predictions_to_poseinterface(
+        predictions_path="fake.csv",
+        video_path="fake.mp4",
+        output_json_parent_dir=tmp_path,
+        sub_id="M01",
+        ses_id="20240101",
+        cam_id="top",
```
lochhh: PR #26 introduces the following fixture
https://github.com/neuroinformatics-unit/poseinterface/blob/07a72900010b759ce459faa97b75eac1282f9dc4/tests/conftest.py#L15-L22

Depending on which gets merged first, we could use something like this here as well
```suggestion
        **sub_ses_cam_ids,
```

lochhh (changes requested):
Thanks @sfmig , I think the decision to use movement for loading predictions and sleap-io for loading annotations makes a lot of sense! Most suggestions are trivial, but the main comments are related to the lack of error-handling in `predictions_to_poseinterface` and tests that exercise the function's various possible behaviours.

niksirbi:
FYI I've now merged https://github.com/neuroinformatics-unit/movement/pull/920 into `movement`, which adds a public [infer_source_software](https://movement.neuroinformatics.dev/dev/api/movement.io.load.infer_source_software.html) function, and a `source_software='auto'` option to `load_dataset`. 

Once we release, this PR can be rebased to directly use `movement` for the 'guessing'.

poseinterface/io.py
niksirbi: I wonder what we should call this file (containing the predictions of the entire video).

It's not part of our spec, but it's obviously useful to cache for extracting more clips, as needed. Suffixing it as `cliplabels.json` has the advantage of signaling the expected format of the .json file, but is a bit awkward because we are not dealing with a `clip` here, but with the whole session video instead.

I wonder if this should just be suffixed with `labels.json`, `videolabels.json`? I will open an issue so we can discuss this matter (and in general, how to handle these 'temporary' predicted labels for the session video).

lochhh: or sessionlabels.json ?

niksirbi: I like `_sessionlabels.json`

niksirbi: I left an extended comment about this in the relevant issue https://github.com/neuroinformatics-unit/poseinterface/issues/10#issuecomment-4281509141

sfmig: I left a relevant comment continuing this conversation in #10 too

niksirbi:
@sfmig when you get the chance to get back to this, you may find it useful to cherry-pick the following commits from PR #49:

-  [`c104ec2` (#49)](https://github.com/neuroinformatics-unit/poseinterface/pull/49/commits/c104ec2fa6b0052878f52206919650ce0f9f9d10) (simplify `predictions_to_poseinterface` using `movement>=0.16` auto source inference) 
-  [`e9aa17e` (#49)](https://github.com/neuroinformatics-unit/poseinterface/pull/49/commits/e9aa17eee20d747570c92de4512fbe2bcba8f9fa) (Rename `cliplabels` to `videolabels` in `predictions_to_poseinterface`)

poseinterface/io.py
sfmig: I don't see a big need to emphasise "session" here, especially since we went for the `videolabels` name rather than `sessionlabels` in the end, so will kept it as `video file` for clarity for now.

tests/test_unit/test_io.py
sfmig: Sounds good. Just for completeness, there is a `video_labels` fixture defined [here](https://github.com/neuroinformatics-unit/poseinterface/blob/42ac536d8cd4eabb4e848730754d6c75fb40a77d/tests/test_unit/test_clips.py#L17). I think initially this fixture was a sibling to it

sfmig: Actually, I kept it as is since there is an overlapping `mock_video` fixture defined [here](https://github.com/neuroinformatics-unit/poseinterface/blob/42ac536d8cd4eabb4e848730754d6c75fb40a77d/tests/test_unit/test_clips.py#L41) which uses `n_frames=10`. I also renamed the factory fixture to `get_mock_video` for clarity. 

We also need to keep MagicMock rather than SimpleNamespace because MagicMock supports magic methods such as `__getitem__` and SimpleNamespace does not. Note we use this e.g. when we do `mock_video[start_frame : start_frame + duration]`

tests/test_unit/test_io.py
```diff
@@ -530,3 +604,129 @@ def test_reencode_video(mock_load_video, mock_save_video, tmp_path):
...
+
+    # Check subfunctions are called
+    mock_guess_source_software.assert_called_once()
+    mock_load_dataset.assert_called_once()
+    mock_load_video.assert_called_once()
+    mock_convert.assert_called_once()
+
+    # Check output file exists with expected name
+    assert result.exists()
+    assert result.name == "sub-M01_ses-20240101_cam-top_videolabels.json"
```
sfmig: reference to comment above

tests/test_unit/test_io.py
sfmig: Thanks for raising this. The intent was to test here the wiring since the underlying function (`_convert_movement_ds_to_videolabels`) was tested separately. But I agree error handling was missing in the implementation and tests. I address the cases you raised below.

> returned file path is as expected and exists

The case for "Returned file path exists / has expected name" was already covered [here](https://github.com/neuroinformatics-unit/poseinterface/pull/45/changes#r3236306835) - or did you mean something else?

> valid/invalid video_path (e.g. video.shape is None)

The following cases related to error handling are now implemented and tested:
* invalid video_path
* video.shape is None

> valid/invalid output_dir (e.g. output dir doesn't exist)

I went for adding `output_dir.mkdir(parents=True, exist_ok=True)` instead, which is more consistent with the implementation in `annotations_to_poseinterface` > `_build_output_json_path`. 

`mkdir` would raise an error if a file (not a dir) exists at that location or if there are permission issues. I set `output_dir=tmp_path / "nested" / "out"` to pass a path that does not exist yet and force that creation (note `tmp_path` is created by pytest at each invocation). 

> when _guess_source_software returns 0, 1, or more matches

I think the test for `_guess_source_software` is no longer necessary because we are doing this via `movement` (via load_dataset > infer_source_software), which is tested there and throws a clear error (see [here](https://movement.neuroinformatics.dev/dev/api/movement.io.load.infer_source_software.html#movement.io.load.infer_source_software)), but do let me know if you think otherwise.

> file content is as expected (a rough check would suffice, since the unit test for _convert_movement_ds_to_cliplabels should check this thoroughly)

Yeah, I understand this is covered in `test_convert_movement_ds_to_videolabels`, but this is a good point I hadn't thought of. I added a check that verifies the output from the convert function is written to disk as a JSON.

sfmig: I agree re integration tests, but maybe it bloats a bit this PR. Maybe they could be derived from the example from @niksirbi [here](https://github.com/neuroinformatics-unit/poseinterface/pull/49)? 

It seems closely aligned, and maybe Claude does a good first draft translating the example to an integration test.

poseinterface/io.py
sfmig: I am relying on `movement` now for this, which I think it addresses all the comments on `_guess_source_software` you raise here, but let me know if something slipped me

sfmig:
Thanks @lochhh and apologies for the delay. I think all comments are addressed, will re-request now.

tests/test_unit/test_io.py
```diff
@@ -530,3 +590,194 @@ def test_reencode_video(mock_load_video, mock_save_video, tmp_path):
...
+        video_path=fake_video,
+        output_dir=tmp_path / "nested" / "out",
+        # (use a nested dir from tmp_path to force creation)
+        **sub_ses_cam_ids,
+    )
+
+    # Check subfunctions are called
+    mock_load_dataset.assert_called_once()
+    mock_load_video.assert_called_once()
+    mock_convert.assert_called_once()
```
lochhh: The wiring will be more robustly tested in integration tests (another PR as you suggested). So there is no need for these asserts. There is still value though in asserting `mock_convert` is called with specific args to ensure we do not accidentally swap `img_h` and `img_w`, e.g.
```python
   # Check video dimensions are correctly wired to the conversion call
    mock_convert.assert_called_once_with(
        ds,
        sub_id=sub_ses_cam_ids["sub_id"],
        ses_id=sub_ses_cam_ids["ses_id"],
        cam_id=sub_ses_cam_ids["cam_id"],
        img_h=480,
        img_w=640,
    )
```
But then again, this is optional as integration tests will cover these. My preference would be to remove these in favour of simplicity.
```suggestion
```

pyproject.toml
```diff
@@ -7,11 +7,14 @@ authors = [
 ]
 description = "A framework for benchmarking pose estimation and point tracking methods on animal beheviour videos."
 readme = "README.md"
-requires-python = ">=3.11.0"
+requires-python = ">=3.12.0"
 dynamic = ["version"]
 
 dependencies = [
+  "jupyter>=1.1.1",
+  "movement>=0.16.0",
```
lochhh:
```suggestion
```
movement dep is listed twice, keep either one (should be fine without min pin)

lochhh (approved):
Thanks again @sfmig for addressing all comments! Pre-approving as there are only two minor things: duplicate movement dependency in pyproject.toml, suggestion to remove assert function calls (wiring will be tested in integration tests).

tests/test_unit/test_io.py
niksirbi: Yeah, the example serves as kind of an integration test for DLC. We should definitely write integration tests for the whole workflow across DLC, SLEAP, LP but that's a separate big chunk of work, best tackled separately.

PR #48, opened 2026-05-11
Patching PR 26

niksirbi:
This PR contains my suggested fixes to the issues surfaced by the latest review of PR #26.


lochhh (approved):
Thank you for uncovering and fixing these issues! 
For the clip format, IDs are assigned as sequential enumerate indices, so  duplicate file_name values would silently produce two images with different IDs but the same filename. I've [added a duplicate filename check in _update_image_ids](https://github.com/neuroinformatics-unit/poseinterface/pull/48/changes/d6de215e91a8db563c8152e8fbf4668b689eb5fd) for this.

PR #40, opened 2026-03-30
Update example for converting DLC project to benchmark

niksirbi:
## Description

**What is this PR**

- [x] Bug fix
- [x] Addition of a new feature
- [ ] Other

**Why is this PR needed?**

We need a working end-to-end example showing how to convert a DeepLabCut (DLC) project to the `poseinterface` benchmark dataset format.
This is both for us to actually use when converting DLC projects, and to provide a reference example to collaborators.

**What does this PR do?**

> [!warning]
> This is a **Frankenstein** branch pulling code from multiple ongoing PRs, including #26 and #39 and #45.
> I created this to facilitate quick-and-dirty conversion. It will need to be rebased to `main` and cleaned up after these other PRs have been merged.

- Adds a lightweight DLC project fixture under `tests/data/dlc/` with 2 sessions, each with a small (100 frames) video, placeholder PNGs (0-byte test files), annotation CSVs, and prediction CSVs (first 100 frames, corresponding to the videos).
- Adds the `convert_dlc_to_benchmark` sphinx-gallery example that runs against the fixture. This example:
	- Uses `annotations_to_poseinterface` (from https://github.com/neuroinformatics-unit/poseinterface/pull/26) for converting DLC annotations to COCO keypoints format.
	- Adds a `frames_to_poseinterface` utility to copy and rename frame images according to filenames in a COCO .json file.
	- Uses `video_to_poseinterface` (from https://github.com/neuroinformatics-unit/poseinterface/pull/39) for re-encoding and renaming videos to our format.
	- Uses the `predictions_to_poseinterface` function from #45 for converting DLC predictions csv to `cliplabels.json` format.
	- Uses `extract_clip` (from https://github.com/neuroinformatics-unit/poseinterface/pull/39) to extract clips from the session video and the corresponding `cliplabels.json` file, and save them in the `Clips/` directory of the benchmark dataset
- Adds a `tree` utility function to display directory structures, with trailing `/` on directory names. This is called several times in the updated sphinx-gallery example to show the structure of the DLC project and the converted benchmark dataset.
- Updates docs config to support sphinx-gallery execution (resolves nbsphinx conflicts, adds matplotlib dependency).

This is a rough schematic of the workflow illustrated by the example:

image

**What is currently missing?**

Waiting for the aforementioned 3 PRs to be reviewed and merged, so this branch can be rebased.

## References

- #26
- #39
- #38 
- #45
- #10 

## How has this PR been tested?

[...]

PR #49, opened 2026-05-12
Update example for converting DLC project to benchmark (v2)

niksirbi:
## Summary

Clean replacement for #40 (Frankenstein branch). Starts fresh from `main`, builds on top of the now-merged #45 (`predictions_to_poseinterface`), and adds the gallery example plus supporting utilities and a lightweight test fixture.

The example walks through the following workflow:

DLC_to_poseinterface_worklow

> [!warning]
> This example specifically documents the DLC-to-poseinterface conversion process. Trying it on other pose estimation software is encouraged, but adaptations will be needed, and the underlying functions are not yet tested on non-DLC inputs. Other source software will be tackled in future PRs.

### What's included

- `frames_to_poseinterface`: copies and renames frame images to match the filenames in a `framelabels.json` (unit tests included).
- Lightweight DLC project fixture under `tests/data/dlc/MouseTopDown-Loukia-2022-09-13/` : 2 sessions, short MP4 videos, placeholder PNG frames, annotation + prediction CSVs. Used by tests and the gallery example.
- `convert_dlc_to_benchmark` sphinx-gallery example: replacing the old `SWC-plusmaze_to_benchmark` example. Runs end-to-end against the bundled fixture and demonstrates the full DLC → benchmark conversion (video, framelabels + frame copy, videolabels) followed by clip extraction.
- `tree` utility: for displaying directory structures, used in the example (unit tests included).
- `movement` pin bumped to `>=0.16.0` (the version introducing the automatic source-software inference, which `predictions_to_poseinterface` now relies on).
- Docs updates:
	- sphinx-gallery execution enabled in `conf.py`
	- API index entries for `tree`, `frames_to_poseinterface`, and `predictions_to_poseinterface`.
	- Docs dependencies: `jupyter`, `matplotlib`.

### Supersedes

- #40

## How was this tested

I used a variant of this example locally to convert 2 sessions from a 'real' dataset (not just the fixture) and inspect the resulting .json files. The real dataset was structured similarly to the fixture included here. 

**Checklist**

- [x] All unit tests pass locally
- [x] Pre-commit  run
- [x] Sphinx-gallery example runs end-to-end (`python examples/convert_dlc_to_benchmark.py`)
- [x] Docs build successfully (`make clean html` from `docs/`)
- [x] CI passes

## How to review

I recommend  primarily reviewing the built example end-to-end, and then diving into the newly introduced `tree`/`frames_to_poseinterface` functions if/when necessary.


poseinterface/io.py
```diff
@@ -466,6 +466,57 @@ def _reencode_video(
     return reencoded_video_path
 
 
+def frames_to_poseinterface(
+    input_dir: Path,
+    output_dir: Path,
+    framelabels_path: Path,
+) -> None:
+    """Copy frame images, renaming them per the COCO JSON filenames.
```
lochhh:
```suggestion
    """Copy and rename frame images to match filenames in COCO JSON.
```

poseinterface/io.py
```diff
@@ -466,6 +466,57 @@ def _reencode_video(
...
+        target_filename = img["file_name"]
+        frame_number = _extract_frame_number(target_filename)
+        if frame_number not in source_frame_map:
+            raise FileNotFoundError(
+                f"No source frame found for frame {frame_number} "
+                f"in {input_dir}"
+            )
+        target_path = output_dir / target_filename
+        if not target_path.exists():
+            shutil.copy2(source_frame_map[frame_number], target_path)
```
lochhh, line 521: Instead of breaking at every missing frame, wdyt about proceeding as usual but warn at the end, so that users could fix the missing frames and then rerun the function? 
```suggestion
    missing_frames = []
    for img in coco_data["images"]:
        target_filename = img["file_name"]
        frame_number = _extract_frame_number(target_filename)
        if frame_number not in source_frame_map:
            missing_frames.append(target_filename)
            continue
        target_path = output_dir / target_filename
        if not target_path.exists():
            shutil.copy2(source_frame_map[frame_number], target_path)

    if missing_frames:
        missing = "\n".join(f"  {f}" for f in missing_frames)
        warnings.warn(
            f"{len(missing_frames)} frame(s) not found in {input_dir} "
            f"and were skipped:\n{missing}",
            UserWarning,
            stacklevel=2,
        )
```

poseinterface/io.py
```diff
@@ -466,6 +466,57 @@ def _reencode_video(
...
+    FileNotFoundError
+        If a source frame cannot be found for a frame number
+        listed in the COCO JSON.
+    """
+    # Build a map from frame number to source image path
+    source_frame_map: dict[int, Path] = {}
+    for img_path in input_dir.glob("*.png"):
+        match = re.search(r"(\d+)", img_path.stem)
+        if match:
+            source_frame_map[int(match.group(1))] = img_path
```
lochhh: we also accept [jpeg format](https://poseinterface.neuroinformatics.dev/benchmark-dataset.html#frames). Should also think about how we want to handle cases when the same frame is available in multiple formats. The suggestion here just implicitly priorities png over jpeg and jpg, by overwriting existing entries for the same frame number. We may also want to either fail fast or exit if the input dir doesn't contain any matching frame files. 
```suggestion
    for ext in ("*.jpg", "*.jpeg", "*.png"):
        for img_path in input_dir.glob(ext):
            match = re.search(r"(\d+)", img_path.stem)
            if match:
                source_frame_map[int(match.group(1))] = img_path

    if not source_frame_map:
        raise FileNotFoundError(
            f"No image files found in {input_dir}"
        )
```

pyproject.toml
```diff
@@ -50,6 +50,7 @@ dev = [
 ]
 docs = [
   "linkify-it-py",
+  "matplotlib",
```
lochhh, line 53: I don't see where this is imported?

tests/data/dlc/MouseTopDown-Loukia-2022-09-13/labeled-data/M708154_EPM_20200317_185651629-converted/CollectedData_Loukia.csv
```diff
@@ -0,0 +1,6 @@
+scorer,,,Loukia,Loukia,Loukia,Loukia,Loukia,Loukia,Loukia,Loukia
+bodyparts,,,snout,snout,left_ear,left_ear,right_ear,right_ear,tailbase,tailbase
```
lochhh, line 2: Since we have 2 CollectedData.csv, shall we use single-index DLC format for one and multi-index for the other? We could potentially remove CollectedData_Pranav.csv and CollecetdData_shailaja.csv from our test data

poseinterface/utils.py
```diff
@@ -0,0 +1,99 @@
...
+        if not level:
+            return
+        contents = sorted(
+            (
+                d
+                for d in dir_path.iterdir()
+                if not (exclude_hidden and d.name.startswith("."))
+            ),
+            key=lambda d: d.name,
+        )
```
lochhh: more readable?
```suggestion
        contents = [
            d
            for d in sorted(dir_path.iterdir(), key=lambda d: d.name)
            if not (exclude_hidden and d.name.startswith("."))
        ]
```

poseinterface/utils.py
```diff
@@ -0,0 +1,99 @@
...
+        if limit_to_directories:
+            contents = [d for d in contents if d.is_dir()]
+        pointers = [tee] * (len(contents) - 1) + [last]
+        for pointer, path in zip(pointers, contents):
```
lochhh:
```suggestion
        last_index = len(contents) - 1
        for index, path in enumerate(contents):
            pointer = last if index == last_index else tee
```

examples/convert_dlc_to_benchmark.py
```diff
@@ -0,0 +1,333 @@
...
+print(tree(source_project_dir / "videos", level=1, exclude_hidden=True))
+
+# %%
+# Each video (ending in ``converted.mp4``) has a companion .csv prediction
+# file. The ``labeled-data`` sub-directories mirror the video names (without
+# .mp4) and contain the sampled frame images (.png) and their annotations
+# (.csv). In real projects you may also find predictions and annotations in
+# .h5 format, as well as filtered prediction files.
+
+print(tree(source_project_dir / "labeled-data", level=2, exclude_hidden=True))
```
lochhh: reordering to:
print(video)
describe video dir
print(labeled-data)
describe labeled-data dir

```suggestion
# %%
# In ``videos/``, each video (ending in ``converted.mp4``) has a companion .csv
# prediction file.

print(tree(source_project_dir / "labeled-data", level=2, exclude_hidden=True))

# %%
# In ``labeled-data/``, the sub-directories mirror the video names (without
# .mp4) and contain the sampled frame images (.png) and their annotations
# (.csv). In real projects you may also find predictions and annotations in
# .h5 format, as well as filtered prediction files.
```

examples/convert_dlc_to_benchmark.py
```diff
@@ -0,0 +1,333 @@
...
+        input_path=source_predictions_path,
+        video_path=source_video_path,
+        output_dir=target_session_dir,
+        **ids,
+    )
+    print(
+        f"\tpredictions: {source_predictions_path.name} -> "
+        f"{sub_ses_cam_prefix}_videolabels.json"
+    )
+    print("Done.\n")
```
lochhh, line 222: also reordering to group code according to description:
1. copy video
2. convert annotations
3. convert predictions

additionally skip (rather than fail) step 2 and/or 3 if the annotation/prediction files do not exist; e.g. there may be projects with annotations but not predictions.

```suggestion
for session in sessions:
    split = session["split"]
    ids = {k: session[k] for k in ["sub_id", "ses_id", "cam_id"]}
    sub_ses_prefix = f"sub-{ids['sub_id']}_ses-{ids['ses_id']}"
    sub_ses_cam_prefix = f"{sub_ses_prefix}_cam-{ids['cam_id']}"
    source_video_path = source_project_dir / "videos" / session["source_video"]
    source_frames_dir = (
        source_project_dir / "labeled-data" / source_video_path.stem
    )
    target_session_dir = (
        benchmark_base_dir / split / project_name / sub_ses_prefix
    )
    target_frames_dir = target_session_dir / "Frames"
    target_frames_dir.mkdir(parents=True, exist_ok=True)

    print(f"Converting session: {split}/{project_name}/{sub_ses_prefix}")
    # Copy the session video, re-encoding to H.264/yuv420p if necessary
    video_to_poseinterface(
        input_video=source_video_path,
        output_video_dir=target_session_dir,
        **ids,
    )
    print(f"\tvideo: {source_video_path.name} -> {sub_ses_cam_prefix}.mp4")

    # Convert DLC annotations to COCO frame labels JSON, then copy the
    # corresponding frame images with standardised poseinterface filenames.
    # In real projects there may be multiple annotation CSVs (e.g. for
    # different labelers); adjust the glob pattern to select the right one.
    source_annotations_path = next(
        source_frames_dir.glob("CollectedData_*.csv"),
        None,
    )
    if source_annotations_path is None:
        print(
            f"\tNo CollectedData CSV found in {source_frames_dir}."
            " Skipping annotations-to-poseinterface conversion."
        )
    else:
        framelabels_path = annotations_to_poseinterface(
            input_path=source_annotations_path,
            output_dir=target_frames_dir,
            format="frame",
            **ids,
        )
        frames_to_poseinterface(
            input_dir=source_frames_dir,
            output_dir=target_frames_dir,
            framelabels_path=framelabels_path,
        )
        print(
            f"\tannotations (+ frame images): {source_annotations_path.name} "
            f"-> {framelabels_path.name}"
        )

    # Convert DLC predictions to COCO video labels JSON for clip extraction.
    # In real projects there may be multiple prediction CSVs (e.g. filtered
    # versions); adjust the glob pattern to select the right one.
    source_predictions_path = next(
        (source_project_dir / "videos").glob(f"{source_video_path.stem}*.csv"),
        None,
    )
    if source_predictions_path is None:
        print(
            f"\tNo prediction CSV found for {source_video_path.stem!r} in "
            f"{source_project_dir / 'videos'}. Skipping predictions-to-"
            "poseinterface conversion."
        )
    else:
        predictions_to_poseinterface(
            input_path=source_predictions_path,
            video_path=source_video_path,
            output_dir=target_session_dir,
            **ids,
        )
        print(
            f"\tpredictions: {source_predictions_path.name} -> "
            f"{sub_ses_cam_prefix}_videolabels.json"
        )
    print("Done.\n")
```

examples/convert_dlc_to_benchmark.py
```diff
@@ -0,0 +1,333 @@
...
+# The resulting benchmark dataset:
+
+print(tree(benchmark_base_dir, level=5))
+
+# %%
+# .. note::
+#
+#    Frame labels (``framelabels.json``) are generated for both splits, but in
+#    the published dataset the ``Test`` split withholds them for evaluation.
+#    See :ref:`benchmark dataset <target-benchmark-dataset>` for details.
```
lochhh: Add (target-dataset-folder-structure)= to the [folder structure](https://poseinterface.neuroinformatics.dev/benchmark-dataset.html#folder-structure) section in the specs?
```suggestion
#    Frame labels (``framelabels.json``) are generated for both splits, but in
#    the **published** dataset the ``Test`` split intentionally omits them for
#    evaluation. See the
#    :ref:`folder structure specification<target-dataset-folder-structure>` for
#    details.
```

examples/convert_dlc_to_benchmark.py
```diff
@@ -0,0 +1,333 @@
...
+
+# %%
+# Extract clips
+# -------------
+# Let's extract short clips from the converted session videos. The resulting
+# clip label files (``cliplabels.json``) can be proof-read and corrected by
+# experts before being shared as part of the benchmark dataset.
+#
+# First, we specify the clip parameters. This step can be run multiple times
+# with different parameters to grow the clip set incrementally.
```
lochhh:
```suggestion
# Clips (short video segments) can be extracted from the converted session
# videos. When the ``videolabels.json`` files are present, the corresponding
# clip label files (``cliplabels.json``) are generated automatically during
# clip extraction.
# These clip label files should then be proof-read and corrected by
# experts before being included in the benchmark dataset.
#
# First, we specify the clip-extraction parameters. This step can be repeated
# with different parameters to incrementally expand the clip set.
```

examples/convert_dlc_to_benchmark.py
```diff
@@ -0,0 +1,333 @@
...
+# %%
+# .. note::
+#
+#    In the published dataset, the ``Train`` split includes all extracted clip
+#    labels (``cliplabels.json``). The ``Test`` split withholds full clip
+#    labels; only clip start labels (``startlabels.json``), derived from each
+#    clip's first frame, are included to support point-tracker evaluation.
+#    The ``videolabels.json`` files generated in the previous section are
+#    intermediate artifacts used for clip extraction, and are never shared.
+#    See :ref:`benchmark dataset <target-benchmark-dataset>` for details.
```
lochhh: Also pointing to the [folder structure](https://poseinterface.neuroinformatics.dev/benchmark-dataset.html#folder-structure) section
```suggestion
#    In the published dataset, the ``Train`` split includes all
#    ``cliplabels.json`` files. The ``Test`` split omits all
#    ``cliplabels.json`` files and instead provides only clip start labels
#    (``startlabels.json``), derived from each clip's first frame,
#    to support point-tracker evaluation.
#    The ``videolabels.json`` files generated in the previous section are
#    intermediate artifacts used for clip extraction, and are never shared.
#    See the :ref:`folder structure specification<target-dataset-folder-\
#    structure>` for details.
```

tests/test_unit/test_io.py
```diff
@@ -439,6 +440,86 @@ def test_pad_integers_to_same_width():
     assert _pad_integers_to_same_width(input) == expected
 
 
+# ---------- Frames to poseinterface ----------------
+
+
+@pytest.fixture
```
lochhh, line 446: We should replace `tests/data/Train/SWC-plusmaze/sub-M708149_ses-20200317/` with the outputs we generate in the example for M708154 and/or M727755 (sample dlc data added in this PR). This would allow us to test conversions end-to-end with actual inputs and the "converted benchmark dataset", besides being the reference/sample, functions also as the expected outputs. That said the current unit tests sufficiently cover the logic. I'm happy for this to be done in a follow-up PR.

tests/test_unit/test_utils.py
```diff
@@ -0,0 +1,109 @@
...
+        assert "0 directories" in result
+
+    def test_files_and_directories(self, tmp_path):
+        """Test tree with a mix of files and directories."""
+        (tmp_path / "a_dir").mkdir()
+        (tmp_path / "b_file.txt").touch()
+
+        result = tree(tmp_path)
+        assert "a_dir/" in result
+        assert "b_file.txt" in result
```
lochhh, line 23: Ensures file names do not end with trailing slash
```suggestion
        assert "b_file.txt" in result
        assert "b_file.txt/" not in result
```

tests/test_unit/test_utils.py
```diff
@@ -0,0 +1,109 @@
...
+        result = tree(tmp_path)
+        lines = result.split("\n")
+        # Root line
+        assert lines[0].endswith("/")
+        # Subdirectory line
+        subdir_line = [line for line in lines if "subdir" in line][0]
+        assert subdir_line.endswith("subdir/")
+        # File line should NOT end with /
+        file_line = [line for line in lines if "file.txt" in line][0]
+        assert not file_line.endswith("/")
```
lochhh: this test becomes redundant if we `assert "b_file.txt/" not in result` in `test_files_and_directories`
```suggestion
```

lochhh (changes requested):
Thanks @niksirbi !

I tried the example with a sample dlc project (that has annotations only) and it broke when it tries to find the predictions file. Otherwise the example works as expected. My suggestion is to skip the conversion(s) in the example if the expected files are not found. 

It's very nice that we now have a sample DLC project in test data. I'm guessing the broken frame files are intentional, but it would be better to have actual frames, since this would allow us to have images with non-zero width and height in the COCO JSON files. As a follow-up PR, we should consider adding the generated benchmark dataset(s) as part of our test data. These could replace the current "reference dataset" `Train/SWC-plusmaze/sub-M708149_ses-20200317/`, giving us a more cohesive and internally consistent set of test fixtures (using the sample DLC project as test inputs and the reference dataset as expected outputs) - our current test data comes from a mix of sample inputs/outputs across different projects, along with multiple ad‑hoc fixtures.

poseinterface/io.py
niksirbi: Thanks for this, I'd forgotten about jpg. 

Implicitly prioritising .png is a good approach in my opinion. I don't expect us to encounter this case—same frame in multiple formats—in any real project.

poseinterface/io.py
niksirbi: Good idea!

pyproject.toml
niksirbi: matplotlib is required by sphinx-gallery's default image scraper, which is enabled even for non-plotting examples. So the docs build fails without it, even if it's not directly used by `poseinterface`'s code or examples.

I suggest we leave it as is. In all likelihood we'll anyway use `matplotlib` somewhere in this project.

tests/data/dlc/MouseTopDown-Loukia-2022-09-13/labeled-data/M708154_EPM_20200317_185651629-converted/CollectedData_Loukia.csv
niksirbi: I'd push back on this suggestion for now. The Loukia CSVs serve a different purpose — they're part of a realistic DLC project fixture, and in a real project we'd expect all .csv files to be of the same type.

The Pranav/Shailaja CSVs are minimal, isolated unit-test fixtures designed specifically to test the two DLC CSV formats.

I think we should consider this together with your other suggestion about harmonising/simplifying fixtures. I'll open an issue to track that.

niksirbi: Noted as part of https://github.com/neuroinformatics-unit/poseinterface/issues/53

tests/test_unit/test_io.py
niksirbi: Good idea. I opened this as issue #53.

niksirbi:
Thanks for the review @lochhh and for re-organising the example in a more logical way.

I've accepted your suggestions, except for the ones about test fixtures (we'll tackle these separately as part of #53).

lochhh (approved):
Thanks! 🚀

PR #51, opened 2026-05-13
Add project overview image to homepage

niksirbi:
I'm proposing to add a project overview image to the homepage.

This is what it would look like (I'm open to re-arranging the order of elements):

image


sfmig:
Nice! I think from top to bottom text >  card > summary image would like a bit more clear

sfmig (approved):
LGTM! thanks @niksirbi

PR #54, opened 2026-05-26
Add installation instructions in README

niksirbi:
## Description

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**

I realised we don't have any sort of installation instructions. This was mostly fine, since the package is not yet on PyPI and probably has no users other than us, but that's about to change. We're planning to share this repo with collaborators soon, who will try to use it to contribute data.

**What does this PR do?**

- Adds some minimal installation instructions in the README, for creating a venv with `uv` and installing the package directly from GitHub
- Fixes a broken link to the contributing guide.

It DOES NOT add an installation guide to the docs; That will follow after we publish the package on PyPI.

## References

N/A

I reproduced the installation steps locally.

## Is this a breaking change?

No.

N/A

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ source .venv/bin/activate  # On macOS and Linux
 .venv\Scripts\activate     # On Windows PowerShell
 ```
 
-Then, install the package directly from GitHub:
+Then, install the package directly from the `main` branch on GitHub:
 
 ```bash
 uv pip install git+https://github.com/neuroinformatics-unit/poseinterface.git@main
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ source .venv/bin/activate  # On macOS and Linux
 Then, install the package directly from GitHub:
 
 ```bash
-uv pip install git+https://github.com/neuroinformatics-unit/poseinterface.git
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -25,6 +25,28 @@ Read the [documentation](https://poseinterface.neuroinformatics.dev/) for more i
...
+```bash
+uv venv --python=3.13
+source .venv/bin/activate  # On macOS and Linux
+.venv\Scripts\activate     # On Windows PowerShell
+```
+
+Then, install the package directly from GitHub:
+
+```bash
+uv pip install git+https://github.com/neuroinformatics-unit/poseinterface.git
```
sfmig: This would install the version at the tip of `main`, right?

I think I would explicitly mention this in the text, and potentially also add `@main` to be extra clear.

```suggestion
uv pip install git+https://github.com/neuroinformatics-unit/poseinterface.git@main
```

sfmig (approved):
LGTM, just a small comment

niksirbi:
Thanks @sfmig! Good call about explicitly specifying the branch. I've added that and will merge as soon as CI passes.

Issue #52, opened 2026-05-14
Automatically add public functions to API docs


sfmig:
**Is your feature request related to a problem? Please describe.**
Right now, we need to manually add the functions we want to expose in the API docs to the [`api_index.rst`](https://github.com/neuroinformatics-unit/poseinterface/blob/main/docs/source/api_index.rst).

**Describe the solution you'd like**
It'd be nice to have the `movement` infrastructure to add them automatically and have one less thing to remember. Maybe this should be added to the NIU cookiecutter?

`movement` script: https://github.com/neuroinformatics-unit/movement/blob/main/docs/make_api.py

**Describe alternatives you've considered**
\

**Additional context**
\

Issue #50, opened 2026-05-13
Proposal: Simplify the project-conversion workflow with a Session-based API


niksirbi:
After the recent work in #49, where I tried to chain together all the individual functions from `poseinterface.io`, I was inspired to come up with a higher-level API for the project-conversion workflow, i.e. **Step 1** of the below schematic.

<img width="4083" height="1354" alt="Image" src="https://github.com/user-attachments/assets/f80b28a7-401a-4881-967c-863a673e7884" />

The workflow currently involved running four per-asset functions (the *primitives*):
- `video_to_poseinterface`
- `annotations_to_poseinterface`
- `frames_to_poseinterface`
- `predictions_to_poseinterface`

It's on the caller to glue them together: discover the source paths, build the per-session output directories under, and call the four primitives in the right order with matching arguments.

The gallery example in #49 makes this concrete: only about a third of its per-session loop body is conversion calls, the rest is path wrangling and output-layout construction. 
The same pattern would apply to anyone writing this kind of pipeline against their own data. 

The proposal is to introduce a generic `Session` dataclass (or maybe `attrs` class for easier built-in validation?) and two orchestration functions in a new `poseinterface/core.py`. The existing primitives in `poseinterface/io.py` can stay mostly unchanged; the new functions would just compose them and write to the spec layout.

## Proposed public API

```python
@dataclass
class Session:
    sub_id: str
    ses_id: str
    cam_id: str                         
    video_path: Path
    split: Literal["Train", "Test"] = "Train" 
    annotations_path: Path | None = None
    frames_source_dir: Path | None = None
    predictions_path: Path | None = None

def convert_session(
    session: Session,
    benchmark_root: Path,
    *,
    project: str,
) -> Path: ...

def convert_project(
    sessions: Iterable[Session],
    benchmark_root: Path,
    *,
    project: str,
) -> None: ...
```

- `Session` carries everything we need to know about one session: its identifiers, which split it should go in, and the source paths. The above session signature is mostly informed by our recent work on DLC projects, we may need to tweak it to generalise well across pose estimation libraries.
- `convert_session` runs the conversion primitives and writes to `<benchmark_root>/<split>/<project>/sub-X_ses-Y/`, following the [benchmark dataset spec](https://poseinterface.neuroinformatics.dev/benchmark-dataset.html). 
- `convert_project` is a thin loop for converting multiple sessions belonging to the same project (sessions self-route to Train/Test via their `split`).

[...]

PR #56, opened 2026-05-29
Add Lightning Pose conversion example

themattinthehatt:
## Description

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**
Current repo only contains conversion utilities for DLC projects.

**What does this PR do?**
Add Lightning Pose conversion: helper functions, small example dataset, and example conversion script.

## References

Reuses much of the conversion utilities developed in #49.

* Ran pre-commit hooks
* Built the documentation and read the auto-generated example page
* Ran pytest

## Is this a breaking change?

No.

Yes; a new example has been added at "examples/convert_lp_to_benchmark.py".

## Checklist:


examples/convert_lp_to_benchmark.py
```diff
@@ -0,0 +1,374 @@
+"""Convert Lightning Pose project to benchmark dataset
+======================================================
+
+Create a ``poseinterface`` benchmark dataset from a Lightning Pose (LP)
+project.
+"""
+
+# sphinx_gallery_thumbnail_path = '_static/project_icon.png'
```
niksirbi: A small cosmetic issue:

This comment should have been auto-hidden in the rendered html but still appears there (I think because it's surrounded by newlines and hence not part of any existing block).

Since this thumbnail is the same as for the DLC example, I think the cleanest solution would be to set this image as the default thumbnail for all examples, and remove the `# sphinx_gallery_thumbnail_path` comment from both the DLC and LP example files.

This can be achieved by adding the following line in `conf.py`:

```python
sphinx_gallery_conf = {
    ...
    "default_thumb_file": "source/_static/project_icon.png",
}
```

The default thumbnail will apply unless overriden by a comment within an example, or unless the example includes matplotlib plots (in that case the ifrst plot becomes the thumbnail, but this can be configured when/as needed).

examples/convert_lp_to_benchmark.py
```diff
@@ -0,0 +1,374 @@
...
+#
+# .. warning::
+#
+#    Lightning Pose saves prediction files to the model output directory,
+#    **not** to the project's ``videos/`` directory.  Before running this
+#    script, move (or copy) each session's prediction CSV — and apply any
+#    manual corrections you have made — into
+#    ``<source_project_dir>/videos/``, named to match the corresponding
+#    video stem (e.g. ``<video_stem>.csv``).
+
```
niksirbi, line 69: Is this necessary? `predictions_to_poseinterface` accepts separate paths for predictions and for video, so should be fine having them in separate folders, unless I'm missing something.

If it's necessary, I wonder how we could help with the copying. For example, we could include a small utility for copying predictions from the model output dir to the videos dir. Similar logic to how we include a `split_lp_collected_data` utility.

Let me know what you think. We could also open an issue to tackle this later after this PR is merged.

examples/convert_lp_to_benchmark.py
```diff
@@ -0,0 +1,374 @@
...
+# Overview
+# --------
+# We'll handle the conversion in two steps:
+#
+# 1. **Convert:** LP project files (videos, frame annotations, and keypoint
+#    predictions) are restructured into the
+#    :ref:`poseinterface benchmark layout <target-benchmark-dataset>`.
+# 2. **Extract clips:** Short video clips and their labels are extracted
+#    from the converted videos and their corresponding keypoint predictions,
+#    ready for expert review.
```
niksirbi, line 38: We could use this opportunity to cross-reference ther other example by adding something like this:

```rst
# 
# The workflow is similar to the one followed in
# :ref:`sphx_glr_auto_examples_convert_dlc_to_benchmark.py`,
# with a few differences explained below.
```

Helpfully, sphinx-Gallery auto-generates a ref label for each example in the pattern `sphx_glr_<gallery_dir>_<filename>`.

niksirbi: This is how that renders:

image

examples/convert_lp_to_benchmark.py
```diff
@@ -0,0 +1,374 @@
...
+        print(
+            f"\tpredictions: {source_predictions_path.name} -> "
+            f"{sub_ses_cam_prefix}_videolabels.json"
+        )
+    print("Done.\n")
+
+# %%
+# The resulting benchmark dataset:
+
+print(tree(benchmark_base_dir, level=5))
```
niksirbi: FYI this output includes the temp `_lp_sessions` folder. We should either mention that again here, or remove that folder before we call `tree` to avoid confusion. 

Another option would be to make the folder hidden from its very creation (`.lp_sessions`), and then set `exclude_hidden=True` in the `tree` call. That's perhaps the easiest solution.

examples/convert_lp_to_benchmark.py
```diff
@@ -0,0 +1,374 @@
...
+            duration=duration,
+        )
+        print(f"Extracted clip: {clip_path.stem}")
+
+
+# %%
+# The resulting benchmark dataset, including the extracted clips and their
+# corresponding labels:
+
+print(tree(benchmark_base_dir, level=5))
```
niksirbi: Same here, `_lp_sessions` is visible in the output.

examples/convert_lp_to_benchmark.py
```diff
@@ -0,0 +1,374 @@
...
+#    (``startlabels.json``), derived from each clip's first frame,
+#    to support point-tracker evaluation.
+#    The ``videolabels.json`` files generated in the previous section are
+#    intermediate artifacts used for clip extraction, and are never shared.
+#    See the :ref:`folder structure specification<target-benchmark-dataset>`
+#    for details.
+
+
+# %%
+# Record provenance (optional)
```
niksirbi, line 321: Many segments, like this one, are repeated varbatim between the 2 examples.

We could get rid of some duplication using RST `.. include::` directives, which work inside Sphinx-Gallery text blocks. We can put shared snippets in a directory like `docs/source/includes/` (which is already excluded from the build via the `**/includes/**` pattern in `conf.py:89`), then reference them from the example comments:

```rst
# %%
# .. include:: /includes/provenance_note.rst
```

The path is resolved relative to the Sphinx source root (with a leading `/`), so it works regardless of where the generated RST ends up. 

That said, I think we *should not* do this, because it impedes discoverability: someone reading the `.py` example file directly won't see the admonition text inline.

This duplication problem will hopefull get better with time, as we add more high-level functions.

poseinterface/io.py
```diff
@@ -723,3 +724,100 @@ def _convert_movement_ds_to_videolabels(
         "annotations": annotations,
         "categories": categories,
     }
+
+
+def split_lp_collected_data(
```
niksirbi, line 729: You have to manually add this function to `docs/source/api_index.rst` for it to appear in the API reference.

We have an existing open issue for automating that processe, see https://github.com/neuroinformatics-unit/poseinterface/issues/52, but for now it's manual.

poseinterface/io.py
```diff
@@ -723,3 +724,100 @@ def _convert_movement_ds_to_videolabels(
...
+    input_path: Path,
+    output_dir: Path,
+) -> dict[str, Path]:
+    """Split a LightningPose project-level CollectedData.csv into
+    per-session CSVs.
+
+    LightningPose stores labels for all sessions in a single project-level
```
niksirbi: In several places in the docstring you use 'LightningPose' without a space, while in the gallery example you use 'lightning Pose' with a space. I assume we want consistency, so I've flagged it here (though there may be other instances as well).

niksirbi (approved):
Thanks for the PR @themattinthehatt !

Special kudos for adding that LP fixture which makes the example run reproducible. 

I'm pre-approving this PR despite the inline comments I left. Consider my comments optional and feel free to address only those you think are important.

The new gallery example duplicates lots of content from the DLC example, but I think that's fine for now. The problem will improve after we introduce higher-level APIs, as discussed in https://github.com/neuroinformatics-unit/poseinterface/issues/50

examples/convert_lp_to_benchmark.py
themattinthehatt: it's not strictly necessary, for now it's more just to streamline this conversion process and make the script work out-of-the-box. perhaps we revisit this once #50 has been addressed?

examples/convert_lp_to_benchmark.py
themattinthehatt: agreed, I like keeping the example as complete as possible for now. once we get a sleap example in here it'll probably be easier to see how to break out some higher-level functions

poseinterface/io.py
themattinthehatt: good catch!

poseinterface/io.py
themattinthehatt: hah thanks for catching that, I blame it on Claude :laughing:

themattinthehatt:
@niksirbi I addressed most of your comments, thanks for the feedback! I'm getting the same doc build failure here as in #57 - is this a `movement` issue?

I double-checked the build docs, everything looks good :ok_hand:

niksirbi:
> @niksirbi I addressed most of your comments, thanks for the feedback! I'm getting the same doc build failure here as in #57 - is this a `movement` issue?

Yes, see my [comment on the other PR](https://github.com/neuroinformatics-unit/poseinterface/pull/57#issuecomment-4683406270).

examples/convert_lp_to_benchmark.py
niksirbi: I'm fine with that.

Issue #63, opened 2026-07-06
Script for copying data from scratch bucket to benchmark bucket


sharmishtaa:
Datasets for the benchmark are first submitted to the scratch bucket where they are saved under the "Train" or "Test" folder. Train datasets will be copied fully to the the benchmark bucket. For test datasets the following two changes need to be made:

- framelabels.json will not be included in the copy to the benchmark bucket.
- For cliplabels.json: the first frame's label will be extracted and saved as a start_label.json. start_label.json will be copied to the benchmark bucket. cliplabels.json will not be included in the copy.

PR #64, opened 2026-07-09
Add functions to extract first frame labels from cliplabels

sharmishtaa:
Adds two new functions to extract only the first frame's labels from cliplabels.json files:
- extract_start_frames_label: processes local files
- extract_start_frames_label_s3: processes files on S3 (requires boto3)

Both functions read *_cliplabels.json files and create corresponding *_startlabels.json files containing only labels for frame id=0.

Before submitting a pull request (PR), please read the [contributing guide](https://github.com/neuroinformatics-unit/.github/blob/main/CONTRIBUTING.md).

Please fill out as much of this template as you can, but if you have any problems or questions, just leave a comment and we will help out :)

## Description

**What is this PR**

- [ ] Bug fix
- [X ] Addition of a new feature
- [ ] Other

**Why is this PR needed?**

This PR addresses the first step in the issue https://github.com/neuroinformatics-unit/poseinterface/issues/63 which is a script for copying benchmark data from the temporary location to the benchmark location. 

**What does this PR do?**
This script adds a function forstripping the cliplabels file and creating a new file that only contains the first label of the frame which will be shared in the benchmark for the Test data.

## References

https://github.com/neuroinformatics-unit/poseinterface/issues/63

Please explain how any new code has been tested, and how you have ensured that no existing functionality has changed.

## Is this a breaking change?

No 

If any features have changed, or have been added. Please explain how the
documentation has been updated.

## Checklist:

- [ X] The code has been tested locally
- [ ] The code has been formatted with [pre-commit](https://pre-commit.com/)

### DIFF
--- a/poseinterface/clips.py
+++ b/poseinterface/clips.py
@@ -11,28 +11,6 @@ import sleap_io as sio
 from . import s3
 
 
-def _suffix_error(name: str, suffix: str) -> str:
-    return f"File must end with '{suffix}', got {name}"
-
-
-def _extract_startlabels_from_dict(clip_labels: dict) -> dict:
-    start_images = [img for img in clip_labels["images"] if img["id"] == 0]
-    if len(start_images) != 1:
-        raise ValueError(
-            "Clip labels must contain exactly one first-frame image with id 0"
-        )
-
-    return {
-        "images": start_images,
-        "annotations": [
-            annot
-            for annot in clip_labels["annotations"]
-            if annot["image_id"] == 0
-        ],
-        "categories": clip_labels["categories"],
-    }
-
-
 def _validate_clip_request(start_frame: int, duration: int) -> None:
     if start_frame < 0:
         raise ValueError(
@@ -302,6 +280,28 @@ def _extract_cliplabels(
         json.dump(clip_labels, f)
 
     return clip_json
+
+
+def _suffix_error(name: str, suffix: str) -> str:
+    return f"File must end with '{suffix}', got {name}"
+
+
+def _extract_startlabels_from_dict(clip_labels: dict) -> dict:
+    start_images = [img for img in clip_labels["images"] if img["id"] == 0]
+    if len(start_images) != 1:
+        raise ValueError(
+            "Clip labels must contain exactly one first-frame image with id 0"
+        )
+
+    return {
+        "images": start_images,
+        "annotations": [
+            annot
+            for annot in clip_labels["annotations"]
+            if annot["image_id"] == 0
+        ],
+        "categories": clip_labels["categories"],
+    }
 
 
 def extract_startlabels(
--- a/poseinterface/clips.py
+++ b/poseinterface/clips.py
@@ -8,6 +8,30 @@ from pathlib import Path
 
 import sleap_io as sio
 
+from . import s3
+
+
+def _suffix_error(name: str, suffix: str) -> str:
+    return f"File must end with '{suffix}', got {name}"
+
+
+def _extract_startlabels_from_dict(clip_labels: dict) -> dict:
+    start_images = [img for img in clip_labels["images"] if img["id"] == 0]
+    if len(start_images) != 1:
+        raise ValueError(
+            "Clip labels must contain exactly one first-frame image with id 0"
+        )
+
+    return {
+        "images": start_images,
+        "annotations": [
+            annot
+            for annot in clip_labels["annotations"]
+            if annot["image_id"] == 0
+        ],
+        "categories": clip_labels["categories"],
+    }
+
 
 def _validate_clip_request(start_frame: int, duration: int) -> None:
     if start_frame < 0:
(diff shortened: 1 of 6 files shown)


sharmishtaa:
I noticed that Matt had already generated startframe_labels.json but SWC and Allen had not, so we need to functionality to do this either locally or on s3. In general I will also add the script for copying data here, so will plan to update pyproject.toml with the boto installation. Just wanted to flag that in case you had other thoughts.

niksirbi:
Thanks for the PR @sharmishtaa ! I will make sure to take a look at this next week when I'm back from FENS.

poseinterface/clips.py
```diff
@@ -8,6 +8,36 @@
 
 import sleap_io as sio
 
+try:
+    import boto3
```
niksirbi: I think we can (and should) just add boto3 as a dependency in `pyproject.toml`. If using `uv`, a simple `uv add boto3` will do that automatically.

poseinterface/clips.py
```diff
@@ -8,6 +8,36 @@
...
+    BOTO3_AVAILABLE = False
+
+
+def _cliplabels_suffix_error(name: str) -> str:
+    return f"Input file must end with '_cliplabels.json', got {name}"
+
+
+def _extract_start_labels(clip_labels: dict) -> dict:
+    start_images = [img for img in clip_labels["images"] if img["id"] == 0]
+    if len(start_images) != 1:
```
niksirbi: Nice sanity check.

poseinterface/clips.py
```diff
@@ -8,6 +8,36 @@
...
+    BOTO3_AVAILABLE = True
+except ImportError:
+    BOTO3_AVAILABLE = False
+
+
+def _cliplabels_suffix_error(name: str) -> str:
+    return f"Input file must end with '_cliplabels.json', got {name}"
+
+
+def _extract_start_labels(clip_labels: dict) -> dict:
```
niksirbi: Good to have this as a private helper. Perhaps we should name it `_extract_startlabels_from_dict` (or `extract_clip_startlabels_from_dict` if you prefer)?

Two reasons:
- `startlabels` (without space or underscore) is how the file is referred to in the space
- the `from_dict` part makes this helper's scope clearly different from the public function (which operates on paths).

poseinterface/clips.py
```diff
@@ -178,6 +208,170 @@ def _extract_cliplabels(
     return clip_json
 
 
+def extract_start_frames_label(cliplabels_path: str | Path) -> Path:
```
niksirbi: In the spirit of my previous comment, we could perhaps name this simply as `extract_startlabels` (or `extract_clip_startlabels` if you prefer).

poseinterface/clips.py
```diff
@@ -178,6 +208,170 @@ def _extract_cliplabels(
...
+        raise ValueError(_cliplabels_suffix_error(cliplabels_path.name))
+
+    # Read the cliplabels file
+    with open(cliplabels_path) as f:
+        clip_labels = json.load(f)
+
+    start_labels = _extract_start_labels(clip_labels)
+
+    # Generate output path by replacing _cliplabels.json with _startlabels.json
+    output_path = cliplabels_path.parent / cliplabels_path.name.replace(
```
niksirbi: Here you assume that the output path will be in tha same dir as the input path.
What do you think of the functions taking two arguments: `cliplabels_path` (same as now) and `output_dir` (or similar?).
The file name should still be derived from the input `cliplabels_path` but in most cases we'd want the the `startlabels` file to go elsewhere (i.e. a different bucket or folder).

But this depends a bit on how we write the script.

1. One option would be for the script to first copy `*.cliplabels.json` files to the 'clean' bucket, then extract `*.startlabels.json` files from them, and finally delete the copied `*.cliplabels.json` files. If that's the plan, current implementation is fine as is.
2. Another option would be to do it all in one step, i.e. have this function accept an input `*.cliplabels.json` file from the scratch bucket and write an output `*startlabels.json` file directly to the clean bucket (nothing needs to be copied or deleted). My suggestion is assuming that option, but I might be overthinking this.

niksirbi: We could also do a hybrid of the two: add `output_dir` but make it optional, with `None` as defauult:
- If `output_dir=None`, save the `startlabels.json` file in the same folder as the input (current behaviour)
- If `output_dir` is supplied, save the `startlabels.json` file there


This keeps all options open for the eventual script.

poseinterface/clips.py
```diff
@@ -178,6 +208,170 @@ def _extract_cliplabels(
...
+    ------
+    ImportError
+        If boto3 is not installed.
+    ValueError
+        If the S3 URI format is invalid or filename does not end with
+        ``_cliplabels.json``.
+    ClientError
+        If there are issues accessing S3 (permissions, file not found, etc.).
+    """
+    if not BOTO3_AVAILABLE:
```
niksirbi: This check won't be necessary once we add the dependency.

poseinterface/clips.py
```diff
@@ -178,6 +208,170 @@ def _extract_cliplabels(
...
+
+    logging.info(
+        f"Extracted start frame labels from {cliplabels_path.name} "
+        f"to {output_path.name}"
+    )
+
+    return output_path
+
+
+def extract_start_frames_label_s3(
```
niksirbi: I expect we'll get to maintain a fair amount of code for interfacing with s3. What do you think of moving the s3 related bits to a new separate module? Could be named `s3.py` for now. This would also make it easier to add a dedicated CLI entrypoint there (should we need it).

poseinterface/clips.py
niksirbi: Also note that all new public functions should be added to `docs/source/api_index.rst` to have them appear in the docs website.
In future, this will happen automatically once we solve https://github.com/neuroinformatics-unit/poseinterface/issues/52 but we haven't gotten there yet.

poseinterface/clips.py
```diff
@@ -178,6 +208,170 @@ def _extract_cliplabels(
...
+        )
+
+    # Extract bucket and key from URI
+    uri_parts = s3_cliplabels_uri[5:].split("/", 1)
+    if len(uri_parts) != 2:
+        raise ValueError(
+            f"Invalid S3 URI format. Expected 's3://bucket/key', "
+            f"got {s3_cliplabels_uri}"
+        )
+    bucket_name, key = uri_parts
```
niksirbi: We may end up re-using S3 URI validation and parsing many places. So perhaps we can already refactor this part into a private helper. But I may be anticipating too much ahead, feel free to ignore for now.

poseinterface/clips.py
niksirbi: If you accept my previous suggestion about functions names, then this one should become `extract_startlabels_s3` (or `extract_clip_startlabels_s3`).

poseinterface/clips.py
niksirbi: Thinking about it some more, there is another possibility: we could require dataset contributors to pre-extract the `startlabels` from `cliplabels` directly at dataset submission. In that case this function is sufficient as is, since we'd only have to move the `startlabels` file from one bucket to another (perhaps that's what you originally had in mind?).

That works but is mildly inelegant, as the contributors essentially supplies the same information (the labels of the first clip frame) in two different formats.

poseinterface/clips.py
```diff
@@ -178,6 +208,170 @@ def _extract_cliplabels(
...
+        logging.info(f"Downloading {s3_cliplabels_uri}")
+        response = s3_client.get_object(Bucket=bucket_name, Key=key)
+        clip_labels = json.loads(response["Body"].read().decode("utf-8"))
+
+        start_labels = _extract_start_labels(clip_labels)
+
+        # Generate output key by replacing _cliplabels.json with
+        # _startlabels.json
+        output_key = key.replace("_cliplabels.json", "_startlabels.json")
+        output_uri = f"s3://{bucket_name}/{output_key}"
```
niksirbi: This has the same issue I highlighed eaerlier about `extract_start_frame_label`: it reads and writes to the same bucket and folder. That may or may not what we want, depending on which workflow we go for.

poseinterface/clips.py
```diff
@@ -178,6 +208,170 @@ def _extract_cliplabels(
...
+        # Download and read the cliplabels file from S3
+        logging.info(f"Downloading {s3_cliplabels_uri}")
+        response = s3_client.get_object(Bucket=bucket_name, Key=key)
+        clip_labels = json.loads(response["Body"].read().decode("utf-8"))
+
+        start_labels = _extract_start_labels(clip_labels)
+
+        # Generate output key by replacing _cliplabels.json with
+        # _startlabels.json
+        output_key = key.replace("_cliplabels.json", "_startlabels.json")
```
niksirbi, line 435: I think this would replace every occurrence (including directory comonents), not just the suffix. Such an occurrence is extremely unlikely, but perhaps the following is safer?

```python
output_key = key.removesuffix("_cliplabels.json") + "_startlabels.json"
```

poseinterface/clips.py
```diff
@@ -178,6 +208,170 @@ def _extract_cliplabels(
...
+        S3 URI of the output ``*_startlabels.json`` file.
+
+    Raises
+    ------
+    ImportError
+        If boto3 is not installed.
+    ValueError
+        If the S3 URI format is invalid or filename does not end with
+        ``_cliplabels.json``.
+    ClientError
```
niksirbi, line 419: Should this be a `FileNotFoundError`? I think the code catches the ClientError and translates it into a `FileNotFoundError`.

niksirbi (changes requested):
Hey @sharmishtaa, finally found the time to take a proper look at this and left comments inline.

I think the main thing that's still unclear in my head is what sort of workflow do we want for generating `startlabels.json` files. I can see at least three alternatives:

1. Data contributors extract `startlabels.json` files themselves and include them at data submission stage. Then the files are already present at the scratch bucket. Your implementation is consistent with that workflow.
2. Data contributors just submit `clipabels.json` files, and our upcoming script will auto-extract `startlabels.json` from them at the time of moving data from one bucket to the other.

The direction this PR will go depends on that key decision.

I personally have a slight preference towards the 2nd option, because I think it's more robust to have our script auto-extract that information, rather than have every data contributor pre-generate `startlabels.json` (essentially duplicating data that's already present in `cliplabels.json`).

Let me know what you think.

Note that tests are currently missing for the two new public functions, but let's first settle on the right workflow/interface before we worry about that.

Let's keep the conversation going here. I'm happy to take another look at this when you need me to.

sharmishtaa:
> Hey @sharmishtaa, finally found the time to take a proper look at this and left comments inline.
> 
> I think the main thing that's still unclear in my head is what sort of workflow do we want for generating `startlabels.json` files. I can see at least three alternatives:
> 
> 1. Data contributors extract `startlabels.json` files themselves and include them at data submission stage. Then the files are already present at the scratch bucket. Your implementation is consistent with that workflow.
> 2. Data contributors just submit `clipabels.json` files, and our upcoming script will auto-extract `startlabels.json` from them at the time of moving data from one bucket to the other.
> 
> The direction this PR will go depends on that key decision.
> 
> I personally have a slight preference towards the 2nd option, because I think it's more robust to have our script auto-extract that information, rather than have every data contributor pre-generate `startlabels.json` (essentially duplicating data that's already present in `cliplabels.json`).
> 
> Let me know what you think.
> 
> Note that tests are currently missing for the two new public functions, but let's first settle on the right workflow/interface before we worry about that.
> 
> Let's keep the conversation going here. I'm happy to take another look at this when you need me to.

I don't think we can expect data contributors to extract the startframe themselves. But what I was unclear about was if we want to leave it as an option - for example Matt had generated his files already. But you bring up a good point about consistency and it will also be a good validation step to ensure that cliplabels.json conforms with what we expect before we copy it. So I'm good with option 2.

sharmishtaa:
Will separate out the s3 utils and will take a look at the other comments later this week.

poseinterface/clips.py
sharmishtaa: > Here you assume that the output path will be in tha same dir as the input path. What do you think of the functions taking two arguments: `cliplabels_path` (same as now) and `output_dir` (or similar?). The file name should still be derived from the input `cliplabels_path` but in most cases we'd want the the `startlabels` file to go elsewhere (i.e. a different bucket or folder).
> 
> But this depends a bit on how we write the script.
> 
> 1. One option would be for the script to first copy `*.cliplabels.json` files to the 'clean' bucket, then extract `*.startlabels.json` files from them, and finally delete the copied `*.cliplabels.json` files. If that's the plan, current implementation is fine as is.
> 2. Another option would be to do it all in one step, i.e. have this function accept an input `*.cliplabels.json` file from the scratch bucket and write an output `*startlabels.json` file directly to the clean bucket (nothing needs to be copied or deleted). My suggestion is assuming that option, but I might be overthinking this.

I had a third option in my head which was how I was going to test it which was to extract the labels in the scratch bucket and then do the copy after, skipping the cliplabels.json. For now, i can just add the option of output_dir so that we can it can work for all options.

poseinterface/clips.py
sharmishtaa: > I expect we'll get to maintain a fair amount of code for interfacing with s3. What do you think of moving the s3 related bits to a new separate module? Could be named `s3.py` for now. This would also make it easier to add a dedicated CLI entrypoint there (should we need it).

thumbs up!

poseinterface/clips.py
sharmishtaa: sounds good.

poseinterface/clips.py
niksirbi: Sounds good.

sharmishtaa:
Hi @niksirbi  - I made the changes to function names, outputs, error catching descriptions, docs, pyproject.toml and separated out the s3 utils. At the moment, I added utilities for just copying json files, but we will need one for generic copying of directories, which can be a separate ticket.

niksirbi (approved):
Thanks for updating the PR @sharmisthaa! 

I added a few commits on top for some finishing touches—fixed linting errors, made sure that typing and docstrings are consistent and added a few more tests. 

I also got rid of the `boto3` availablity check: that would be useful if `boto3` was an optional extra dependency, but we can safely assume its availability since it's now a core dependency.

The only significant behaviour I added was to [also validate fie suffixes for output_path/uri](https://github.com/neuroinformatics-unit/poseinterface/pull/64/commits/34e77e686e7d0eeab1a2610aef4c0c6e2cfe375b): basically, output paths (and URIs) are also validated (if provided) to ensure that they end with `startlabels.json` (as the schema stipulates).

The only remaining task is to resolve the merge conflicts by rebasing. I am happy to take care of that; just wanted to check in that you are fine with the changes I made (any of them can be easily undone) before I undertake the rebasing (which will alter the git history).

sharmishtaa:
Looks good to me. Please go ahead!

PR #67, opened 2026-07-16
Publish to PyPI via pypa action directly

niksirbi:
## Description

**What is this PR**

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**

`upload_pypi@v2` failed at the publish step with a Docker error before it ever contacted PyPI. I hit this while releasing `poseinterface` v0.1.0, so I published that release manually as a workaround.

**Failed job:** https://github.com/neuroinformatics-unit/poseinterface/actions/runs/29440557606/job/87439010800

We had hit this issue before in other projects: see https://github.com/neuroinformatics-unit/actions/issues/76.

**What does this PR do?**

All of our packages now use the `pypa/gh-action-pypi-publish` action directly, to avoid that exact failure.
This PR aligns the PyPI release step with [that of `movement`](https://github.com/neuroinformatics-unit/movement/blob/main/.github/workflows/test_and_deploy.yml), which we know works.

The faulty workflow was inherited from the python-cookiecutter (see existing https://github.com/neuroinformatics-unit/python-cookiecutter/issues/173).

## References

existing actions issue: https://github.com/neuroinformatics-unit/actions/issues/76
existing cookiecutter issue: https://github.com/neuroinformatics-unit/python-cookiecutter/issues/173

The real test will be the next release, but we know the same workflow is functional for dozens of other packages.

## Is this a breaking change?

No.

No.

## Checklist:

- [ ] The code has been tested locally

---

## UPDATED: 2026-07-27

Following @lochhh's [review suggestion](https://github.com/neuroinformatics-unit/poseinterface/pull/67#pullrequestreview-4737291555), this PR now also switches PyPI releases to [Trusted Publishing](https://docs.pypi.org/trusted-publishers/), rather than only swapping out the broken `upload_pypi@v2`.

**What changed since the original description:**

- **On PyPI:** `poseinterface` is configured with a trusted publisher pointing at this repo's `test_and_deploy.yml`, with `pypi` as the environment name.
- **In the workflow:** `upload_all` declares the `pypi` environment and `id-token: write`; the `user`/`password` inputs are gone. No long-lived API token is involved in publishing anymore.
- **In repository Settings > Environments:** Added a `pypi` environment with a deployment tag rule to only allow tags matching `v*`
- **In the docs:** `contributing.md` now states that release tags follow the `vX.Y.Z` format.

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -56,6 +56,11 @@ jobs:
     needs: [build_sdist_wheels]
     runs-on: ubuntu-latest
     steps:
-    - uses: neuroinformatics-unit/actions/upload_pypi@v2
+    - uses: actions/download-artifact@v8
       with:
-        secret-pypi-key: ${{ secrets.TWINE_API_KEY }}
+        name: artifact
+        path: dist
+    - uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        user: __token__
+        password: ${{ secrets.TWINE_API_KEY }}
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -55,12 +55,14 @@ jobs:
     name: Publish build distributions
     needs: [build_sdist_wheels]
     runs-on: ubuntu-latest
+    environment: # use Trusted Publisher route
+      name: pypi
+      url: https://pypi.org/p/poseinterface
+    permissions:
+      id-token: write
     steps:
     - uses: actions/download-artifact@v8
       with:
         name: artifact
         path: dist
     - uses: pypa/gh-action-pypi-publish@release/v1
-      with:
-        user: __token__
-        password: ${{ secrets.TWINE_API_KEY }}
--- a/docs/source/contributing.md
+++ b/docs/source/contributing.md
@@ -121,6 +121,7 @@ A GitHub actions workflow (`.github/workflows/test_and_deploy.yml`) has been set
 * Linting checks (pre-commit).
 * Testing (only if linting checks pass)
 * Release to PyPI (only if a git tag is present and if tests pass).
+  Release tags must follow the `vX.Y.Z` format, e.g. `v0.1.0`.
 
 Another workflow (`.github/workflows/docs_build_and_deploy.yml`) is set up to build and deploy the documentation to GitHub pages on each push to `main` and on releases (i.e. when a git tag is present).
 


niksirbi:
Note: I've already set up `TWINE_API_KEY` in repo secrets, so following this fix, the next releases should work automactically.

lochhh (approved):
Thanks for this @niksirbi ! Changes LGTM. I wonder though, if this is a good opportunity to try switching to the recommended [trusted publishing](https://docs.pypi.org/trusted-publishers/adding-a-publisher/) and drop `user/password` entirely from `test_and_deploy.yml`:
```yaml
upload_all:
    name: Publish build distributions
    needs: [build_sdist_wheels]
    runs-on: ubuntu-latest
    environment:
      name: pypi
      url: https://pypi.org/p/poseinterface
    permissions:
      id-token: write  # mandatory, trusted publishing
    steps:
    - uses: actions/download-artifact@v8
      with:
        name: artifact
        path: dist
    - uses: pypa/gh-action-pypi-publish@release/v1
```

niksirbi:
Thanks @lochhh. Trusted Publishing is something we haven't yet set up in any of our orgs/repos, but I'm sympathetic to the idea of `poseinterface` being the testing ground for it.

The caveat being that, assuming this works, we'll then have to roll this out across our organisations and repos, or else run the risk of publishing workflows diverging (and hence become harder for us to manage long-term).

Given that we have to do this sooner or later, I'm all for starting here. What do you think @adamltyson ?

niksirbi:
Note @K-meech has already proposed switching to trusted environments on our cookiecutter: https://github.com/neuroinformatics-unit/python-cookiecutter/issues/173.

lochhh:
I don't see this as a caveat but rather us moving away from long-lived API tokens to [more secure](https://docs.pypi.org/trusted-publishers/security-model/) short-lived ones. We could also then consider retiring `upload_pypi` and close https://github.com/neuroinformatics-unit/actions/issues/76 since it just wraps 2 steps (download + publish).

niksirbi:
> I don't see this as a caveat but rather us moving away from long-lived API tokens to [more secure](https://docs.pypi.org/trusted-publishers/security-model/) short-lived ones. We could also then consider retiring `upload_pypi` and close [neuroinformatics-unit/actions#76](https://github.com/neuroinformatics-unit/actions/issues/76) since it just wraps 2 steps (download + publish).

I agree, and I would second getting rid of the `upload_pypi` action. It doesn't work as is anyway and it has led me astray several times.

niksirbi:
Tagging also @alessandrofelder, in case he's interested in including this as part of BrainGlobe's general infrastructure overhaul this summer.

niksirbi:
I've now set up Trusted Publishing, in two parts:

- **On PyPI:** configured `poseinterface` with a trusted publisher pointing at this repo's `test_and_deploy.yml`, with `pypi` as the environment name.
- **Updated the workflow** as per @lochhh's suggestion above.I also added a line to the contributing guide noting that release tags follow `vX.Y.Z`.

One open question: I've left the `pypi` environment to be auto-created by GitHub on the first tagged release, which means it starts with no protection rules. We could tighten it by adding  a **deployment tag policy**, restricting the environment to `v*` refs. Right now the only thing gating publishing is `if: github.ref_type == 'tag'` in the workflow file. An environment rule lives in repo settings instead, so it holds independently. Do you think that is worth it here? I personally lean toward yes.

Next steps:

1. Merge this PR
2. Make a new release to confirm the trusted publishing route works
3. Remove unused `TWINE_API_KEY` from repo settings and PyPI
4. Update the cookiecutter to the new route, thereby closing neuroinformatics-unit/python-cookiecutter#173
5. Delete our custom `upload_pypi` action, thereby closing neuroinformatics-unit/actions#76
6. Do the same for `movement` and notify maintainers of other NIU/BrainGlobe packages in case they want to do the same.

lochhh:
> One open question: I've left the `pypi` environment to be auto-created by GitHub on the first tagged release, which means it starts with no protection rules. We could tighten it by adding a **deployment tag policy**, restricting the environment to `v*` refs. Right now the only thing gating publishing is `if: github.ref_type == 'tag'` in the workflow file. An environment rule lives in repo settings instead, so it holds independently. Do you think that is worth it here? I personally lean toward yes.

Thanks @niksirbi ! this is definitely within the scope of this PR.
I don't have access to the settings. How is the environment rule configured? (I looked in movement but can't see this). I suppose we would want to have the same across all repos right?

niksirbi:
Alright, I've also now configured the **deployment tag policy**, using the following commands (sharing here, in case we want to re-use them in other repos):

```sh
REPO=neuroinformatics-unit/poseinterface

# 1. Create the environment with custom ref policies enabled
gh api -X PUT "repos/$REPO/environments/pypi" --input - <<'EOF'
{"deployment_branch_policy":{"protected_branches":false,"custom_branch_policies":true}}
EOF

# 2. Restrict it to v* tags
gh api -X POST "repos/$REPO/environments/pypi/deployment-branch-policies" --input - <<'EOF'
{"name":"v*","type":"tag"}
EOF
```

This can be also done via GitHub repo setting (in the "Environments" section), but probably requires one to be a repo admin.

niksirbi:
## UPDATE

This worked! I've successfully made a new release on PyPI using the trusted publishing route. I've also removed the stale API token. Now I'll move through the rest of the checklist.

PR #73, opened 2026-07-30
S3 copy utils

sharmishtaa:
Before submitting a pull request (PR), please read the [contributing guide](https://github.com/neuroinformatics-unit/.github/blob/main/CONTRIBUTING.md).

Please fill out as much of this template as you can, but if you have any problems or questions, just leave a comment and we will help out :)

## Description

**What is this PR**

- [ ] Bug fix
- [X] Addition of a new feature
- [ ] Other

**Why is this PR needed?**
Copying files to the benchmark dataset

**What does this PR do?**
Adds some utils for copying data and some example scripts that use it.

## References

Please reference any existing issues/PRs that relate to this PR.

Please explain how any new code has been tested, and how you have ensured that no existing functionality has changed.

## Is this a breaking change?

No

If any features have changed, or have been added. Please explain how the
documentation has been updated.

## Checklist:

- [X ] The code has been tested locally
- [X ] Tests have been added to cover all new functionality
- [ ] The code has been formatted with [pre-commit](https://pre-commit.com/)


poseinterface/s3.py
```diff
@@ -121,3 +132,358 @@ def upload_json_to_s3(
...
+        except ClientError as e:
+            logging.error(
+                "Failed to delete objects from s3://%s: %s",
+                bucket_name,
+                e,
+            )
+            raise
+
+
+def copy_s3_folder(
```
niksirbi: The module-level docstring says:

> All functions in this module create a new boto3 Session and S3 client on each invocation.

While this is an established and documented pattern by now, I wonder if it will run into bottlenecks for this bulk-copy utility?

If we wanted to do the whole folder copy in one client call, could we create it once in this function and then pass it as an optional argument to the copy/delete helpers being called?

I have zero experience with S3, so feel free to ignore my comment if this is a non-issue.

examples/copy_s3_folder.py
```diff
@@ -0,0 +1,183 @@
...
+        if success:
+            logging.info("=" * 60)
+            logging.info("✓ Copy completed successfully!")
+            logging.info(f"  Total files copied: {len(copied_keys)}")
+            logging.info("=" * 60)
+            return 0
+        else:
+            logging.error("Copy failed - changes were rolled back")
+            return 1
+
```
niksirbi: The `else:` case here is unreachable, because the `copy_s3_folder` utility always returns `success=True` unless it raises an error.

I think we could drop the boolean `success` flag entirely, and rely just on `copied_keys`. But heads up that this change would necessitate changes in the utility itself, in this script, and some tests in `tests/test_s3.py`.

examples/extract_startlabels_s3.py
```diff
@@ -0,0 +1,134 @@
...
+Extract with automatic output naming:
+    python extract_startlabels.py s3://bucket/path/video_cliplabels.json
+
+Specify custom output location:
+    python extract_startlabels.py s3://bucket/path/video_cliplabels.json \\
+        s3://bucket/output/video_startlabels.json
+
+Use specific AWS profile:
+    python extract_startlabels.py s3://bucket/path/video_cliplabels.json \\
+        --profile my-profile
```
niksirbi: These example use the wrong script name, it should be `python extract_starlabels_s3.py`

examples/copy_s3_folder.py
```diff
@@ -0,0 +1,183 @@
...
+    exclude_filter = None
+    if args.exclude:
+        exclude_filter = create_filename_exclude_filter(args.exclude)
+
+    # Perform the copy
+    try:
+        if args.dry_run:
+            logging.info("Dry run mode - would perform copy operation here")
+            logging.info("No files were actually copied")
+            return 0
```
niksirbi: Having a dry-run option could be really useful, but this here doesn't seem to do anything other than logging these 2 messages?

I would expect a dry-run option to return/print a list of objects that would have been copied.

I think we could implement it along these lines, as we already have all the necessary pieces in place:

```python
if args.dry_run:
    source_objects = list_s3_objects(source_bucket, source_prefix, args.profile)
    for obj in source_objects:
        rel = obj["Key"][len(source_prefix):]
        if obj["Key"] == source_prefix:
            continue
        if exclude_filter and exclude_filter(rel):
            logging.info(f"  SKIP  {rel}")
        else:
            logging.info(f"  COPY  s3://{dest_bucket}/{dest_prefix}{rel}")
    return 0
```

niksirbi: Or perhaps it would be even cleaner to make the `copy_s3_folder` utility itself dry-run aware, i.e. add a `dry_run: bool = False` parameter that runs the same enumeration/filter loop, but logs the results instead of calling `copy_s3_object`. Then this CLI script would just pass the `--dry-run` option through to the underlaying utility.

poseinterface/s3.py
```diff
@@ -1,7 +1,18 @@
-"""S3 utilities for poseinterface."""
+"""S3 utilities for poseinterface.
```
niksirbi: This PR adds several new public functions. Let's also make sure that they appear in API docs.

You can add the new 5 funcs to `docs/source/api_index.rst`, Sphinx will do the rest automatically.

examples/copy_s3_folder.py
niksirbi: That last alternative (make the `copy_s3_folder` utility dry-run aware) is even more attractive if you take my suggestion about moving the CLI code into the package modules.

poseinterface/s3.py
```diff
@@ -121,3 +132,358 @@ def upload_json_to_s3(
...
+        # Regex pattern
+        if pattern.startswith("regex:"):
+            regex_pattern = pattern[6:]
+            return bool(re.match(regex_pattern, path))
+
+        # Glob-style pattern - convert to regex
+        # Escape special regex chars except * and ?
+        regex = re.escape(pattern)
+        regex = regex.replace(r"\*", ".*")
+        regex = regex.replace(r"\?", ".")
```
niksirbi: Maybe it's worth adding a note about this behaviour in the docstring. We want this, but it's slightly surprising.

```md
Unlike shell globs, * and ? also match /. So, for example,
*.json matches JSON files in all subdirectories, not just the top level.
```

poseinterface/s3.py
```diff
@@ -121,3 +132,358 @@ def upload_json_to_s3(
...
+    >>> filter_fn("Train/sample.json")  # False - not excluded
+    >>> filter_fn("data.json")  # False - not excluded
+    """
+
+    def matches_pattern(path: str, pattern: str) -> bool:
+        """Check if path matches a single pattern."""
+        # Regex pattern
+        if pattern.startswith("regex:"):
+            regex_pattern = pattern[6:]
+            return bool(re.match(regex_pattern, path))
```
niksirbi: There is a minor inconsistency here: regex patterns are matched with re.match, anchored at the start only (essentially a prefix match), whereas glob path anchors are both ends.

I'd either mention this in the docstring, or switch this line to `re.fullmatch` so regex behaves like the glob path (both ends anchored).

niksirbi (changes requested):
Thanks for this @sharmishtaa and sorry for taking so long to review this. 
The functions and CLI scripts you added are great and thoroughly tested. 

I have one major comment which I explain here. I've left other minor/optional points as inline comments.

Most should be easy to fix before merging.

## The 3 new CLI scripts are mixed with sphinx-gallery examples

`examples/` is a sphinx-gallery source directory and every `.py` file within it is interpreted as an example sphinx-gallery will try to build and eventually publish on [the website](https://poseinterface.neuroinformatics.dev/auto_examples/index.html).

The 3 scripts you added are argparse CLI tools and do not follow the sphinx-gallery format; hence Sphinx fails to build them, causing the CI error.

My recommendation would be to follow the pattern we already have for the `extract-clips` command, i.e. move these 3 scripts inside package module and declare them as console entry points in `pyproject.toml`, under [`project.scripts`](https://github.com/neuroinformatics-unit/poseinterface/blob/d7f27ab77440e870be59da9193de547499a4386d/pyproject.toml#L66-L67). For example, the existing `extract-clips` command is declared there as `extract-clips = "poseinterface.clips:wrapper"`.

This is a possible mapping that would work in this case:

| Script                           | Wraps                          | Suggested package module home | Suggested entry point                                      |
| -------------------------------- | ------------------------------ | ----------------------------- | ---------------------------------------------------------- |
| `copy_s3_folder.py`              | `s3.copy_s3_folder`            | `poseinterface/s3.py`         | `s3-copy-folder`                                           |
| `extract_startlabels_s3.py`      | `clips.extract_startlabels_s3` | `poseinterface/clips.py`      | `s3-extract-startlabels`                                   |
| `extract_startlabels_s3_list.py` | loops `extract_startlabels_s3` | `poseinterface/clips.py`      | `s3-extract-startlabels-batch` |
|                                  |                                |                               |                                                            |

Moving the CLI commands into the package and registering the `project.scripts` entries would fix the docs CI build, and make the scripts installable. They could then be used by directly invoking the command name (e.g. `s3-copy-folder`), without needing `sys.path` workarounds.

Note that I propose to prefix the CLI commands with `s3-`. I think this would make it clearer that these particular commands change things on s3, as opposed to `extract-clips` which is a command run locally by data contributors. On the other hand, maybe it's more important to have the CLI entry-point names match the corresponding wrapped Python functions (which is what you were going for, I think)? I'm happy with you deciding on what's best here.

Moreover, with my proposal `poseinterface/clips.py` will now hold 3 entry-points: the existing `extract-clips` and the two new `s3_extract-startlabels*` commands. We should give their wrappers distinct names (as opposed to the generic existing `wrapper` function in `poseinterface/clips.py`).

